PR #8, opened 2024-01-20
Automation controller billing feature v3

Ladas:
Gather and ship billing data to console.redhat.com with automatically collecting gap, by storing a last collected
    timestamp and always collecting from that last successfully collected timestamp


slemrmartin (approved):
:+1: it looks goood

PR #41, opened 2025-01-17
add data and complete test for ccspv2 file

ShaiahWren:
[AAP-38383]

This pr adds a test in `metrics-utility/metrics-utility/test/ccspv_reports/test_CCSPv2.py`
It opens the xlsx file and verifies spreadsheet tabs (sheet names) and columns for each. 

## Setup and testing
- Activate the awx virtual environment:
`docker exec -it tools_awx_1 /bin/sh -c "
    cd awx-dev/metrics-utility &&
    . /var/lib/awx/venv/awx/bin/activate &&
   exec /bin/sh
 "`
- From the /metrics-utility dir, run  `SETUPTOOLS_USE_DISTUTILS=true pytest -s`

### Expected Results
All tests should pass

### Useful code explanations
- `env_vars` : a dictionary of environment variables to be used when calling the build_report command.
- `file_path` :  path to the ccspv file that the build_report command will generate.
- `EXPECTED_SHEETS` : a dictionary where the keys, such as "Usage Reporting", represent the tabs (sheet names), and the values as list represent the columns each sheet name is expected to have. 
-  `cleanup()` fixture that removes the xlsx file before and after the test has completed. 
- `test_command()` : the main test body that calls the build_report command, asserts a 0 return code for success, and calls helper functions -- - 
-  `validate_sheet_tab_names()` : reads the file actual sheet names with `pd.ExcelFile` and compares that list of sheet names with the keys in EXPECTED_SHEETS.
-  `validate_sheet_columns()` : again, reads the actual columns using pd.read_excel, loops through each sheet name (key) and column list (value) in EXPECTED_SHEETS and compares them to each other.


MilanPospisil:
Great Job!!!

Now according with Martin, we will be running tests inside the container for now.

So the pytest will run inside awx container using:

exec -it tools_awx_1 /bin/sh -c "
   cd awx-dev/metrics-utility &&
   . /var/lib/awx/venv/awx/bin/activate &&
   exec /bin/sh
   pytest
"

This will run your test, so you dont need to include the exec ... into the PR. Its only important for running the local test.

This exec will run tests inside AWX, so we can run actual CCSPv2 test (because that test needs AWX and must be run inside the container). You will have to set environmental variables and test running the shell command. Do not call function for building report directly in python, but rather really call it as shell command, because we need to test it as command - user will call command, not python function.

First remove the xlsx file that can be potentialy there from previous tests (use some python library, first validate if the file is there and if yes, delete it)

Then we run the command:

import subprocess

result = subprocess.run(
    ['python manage.py build_report --month=2024-12'],      
    text=True,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    env={'VAR1': 'val1', 'VAR2':'val2'}  
)

{'VAR1': 'val1', 'VAR2':'val2'}  Here there should be all env variables from readme.md that command requires

The command will generate the xlsx file.

Also assert the result.returncode to 0. 0 means success, anything else is error, in that case, test should fail.

And now you can assert the contents of the xlsx file as you already have in your PR :)

Dont forget to remove xlsx file from the PR, not needed now :)

Again, great work Shaiah :-) 

Also - you are testing data from 2025-01, this may be problematic right now, because you dont have them all, new data will arrive for this month and if anyone will gather them locally, the report will analyze more data than you supposed to and test will fail.

If possible, test only data from 2024-12 and remove 2025-01 from the test - this means the values for assertion will be different.

You are also using data from shipped_data folder, you should use testing folder, so instead of this:


export METRICS_UTILITY_SHIP_PATH=/awx_devel/awx-dev/metrics-utility/shipped_data/billing

do this:
export METRICS_UTILITY_SHIP_PATH=/awx_devel/awx-dev/metrics-utility/metrics_utility/test/test_data/

And move your testing data from shipping folder to test folder. Also update report path in the test.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,88 @@
...
+        actual_columns == expected_columns
+    ), f"Column names do not match for sheet: {sheet_name}"
+
+
+def test_sheet_tab_names():
+    """Test the sheet names in the Excel file."""
+    excel_data = pd.ExcelFile(file_path)
+    assert excel_data.sheet_names == list(
+        EXPECTED_SHEETS.keys()
+    ), "Sheet names do not match."
```
dhaustein: Question: Isn't the excel sheet the _output_ of the utility? That is, if we are to test the metrics utility cli for billing processing, we need to feed it billing data (here I guess that's the army of tar archives) and then assert the output excel table is what we expect it to be, right? I think this is what Milan suggests above.

MilanPospisil: Yes, Shaiah was instructed to test existing xlsx file as first step. Next step is to call the report generating itself.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
...
+            print(f"Expected columns (formatted): {expected_columns}")
+
+        assert (
+            actual_columns == expected_columns
+        ), f"Column names do not match for sheet: {sheet_name}"
+
+
+def cleanup():
+       if os.path.exists(file_path):
+        os.remove(file_path)
```
dhaustein: suggestion: If using pytest, this could be turned into a pytest fixture that does the cleanup in both setup and teardown

https://docs.pytest.org/en/latest/how-to/fixtures.html#yield-fixtures-recommended

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
+import pandas as pd
+import os
+import subprocess
+import sys
+
```
dhaustein, line 8: suggestion: We should start using `pytest` framework for new tests. Unless of course I missed some decision not to do that.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
...
+    "Usage by organizations": [
+        "Organization name",
+        "Job runs",
+        "Unique managed nodes\nautomated",
+        "Non-unique managed\nnodes automated",
+        "Number of task\nruns",
+    ],
+}
+
+def test_command():
```
dhaustein: suggestion: Please add a doscstring to this test case. :slightly_smiling_face:

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
...
+        expected_columns = [normalize_column(col) for col in expected_columns]
+
+        if actual_columns != expected_columns:
+            print(f"Mismatch for sheet: {sheet_name}")
+            print(f"Actual columns (formatted): {actual_columns}")
+            print(f"Expected columns (formatted): {expected_columns}")
+
+        assert (
+            actual_columns == expected_columns
+        ), f"Column names do not match for sheet: {sheet_name}"
```
dhaustein: nitpick: And this is really a big nitpick. Auxiliary functions and test fixtures usually go before the main body of tests in a file.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
...
+    ],
+}
+
+def test_command():
+    print("Test running")
+
+    cleanup()
+
+    python_executable = sys.executable
+    result = subprocess.run(
```
dhaustein: note: In the future we might look into switching to calling Django's `callCommand()` but that is not needed right now, I just don't want to be alone in remembering that :grin: 

https://docs.djangoproject.com/en/4.2/ref/django-admin/#running-management-commands-from-your-code

dhaustein (commented):
Looks pretty good to me, though I think we can use this opportunity to pickup using pytest

WDYT @himdel @MilanPospisil ?

EDIT: I gave an approve as my comments are all non-blocking and I agree with Milan that we can do the extra stuff in a follow-up PR

metrics_utility/test/ccspv_reports/test_CCSPv2.py
MilanPospisil: We for sure can, but since this have to be ideally merged today (end of sprint), what about to do the rewrite in follow up PR?

metrics_utility/test/ccspv_reports/test_CCSPv2.py
```diff
@@ -0,0 +1,111 @@
...
+EXPECTED_SHEETS = {
+    "Usage Reporting": [
+        "CCSP NA Direct Reporting Template",
+        "Unnamed: 1",
+        "Unnamed: 2",
+        "Unnamed: 3",
+        "Unnamed: 4",
+        "Unnamed: 5",
+        "Unnamed: 6",
+        "Updated: Jan 21, 2025",
```
MilanPospisil: This one must be changed. Updated should match current date, it should not be fixed. Today is for example Jan 22, 2025 and test fails :-)

MilanPospisil (approved):
LGTM :+1: 

But first, address pls at least my comment, so the tests will work for different days than Jan 21 2025 :-)

Since the PR have to be merged today (end of sprint), lets address the changes in follow up PR.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
MilanPospisil: Lets do this in follow up PR.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
dhaustein: I have no objections to that. But it should be a straightforward change. Import pytest and use a fixture decorator with yield

I will leave it up to @ShaiahWren  to decide.

dhaustein (approved):

metrics_utility/test/ccspv_reports/test_CCSPv2.py
art-tapin: I :100: agree wiht @dhaustein on using pytest for our new tests is a great step forward. Additionally, in the future integrating `tox` would improve our testing workflow. It automates environment setup, managing dependencies across different setups, it is a great tool to ensure that our tests run both locally and in CI and helps with "it works on my machine" issues. I worked with tox in my previous team and it is quite a neat tool. As @MilanPospisil  suggested, we can address this separately as a team.

https://tox.wiki/en/latest/
https://christophergs.com/python/2020/04/12/python-tox-why-use-it-and-tutorial/

ShaiahWren:
Thanks for your input everyone. I will work on addressing your suggestions this morning.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
ShaiahWren: Easily fixed:)

metrics_utility/test/ccspv_reports/test_CCSPv2.py
ShaiahWren: @dhaustein I implemented the cleanup as a pytest fixture.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
ShaiahWren: @dhaustein, I added pytest.

metrics_utility/test/ccspv_reports/test_CCSPv2.py
ShaiahWren: Added the docstring.

ShaiahWren:
Linters finally pass. Merging!

PR #42, opened 2025-01-17
add pre-commit with ruff

ShaiahWren:
[[AAP-38768](https://issues.redhat.com/browse/AAP-38768)]

## Description
This work ensures that each time a developer attempts a commit, 
- a pre-commit hook with uv+ruff for linting and formatting is run
- in order to arrest the commit if linting/formatting discrepancies are detected and provide useful feedback in the output log.

##  Set-up
- in /metrics-utility root dir, run ` uv sync` & `uv run ruff / uvx ruff` to manage the uv virtual env and add the dependencies.
- Ensure deps have been installed with checks such as `ruff --version; pytest --version; django-admin --version; pre-commit --version;`
- If deps are missing try `uv sync` again 

### Testing
- create a small test file, such as "test.py", and type any incorrectly formatted or linted python. 
- add the file and attempt to commit the change

### Expected Results
Commit should _not_ complete; it will direct you some output that explains what the pre-commit-with-uv+ruff found.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.9.2
+    hooks:
+      - id: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
+  - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.9.2
     hooks:
       - id: ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,11 +2,17 @@
 requires = ["setuptools>=45", "setuptools_scm[toml]>=6.0"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "metrics-utility"
+version = "0.1.0"
+description = "A metrics utility for Ansible"
+dependencies = [
+    "ruff",
+    "pytest",
+    "Django>=4.0,<5.0",
+    "pre-commit",
+]
+
 # Do not uncomment the line below. We need to be able to override the version via a file, and this
 # causes the "version" key in setup.cfg to be ignored.
 # [tool.setuptools_scm]
-
-[tool.black]
-line-length = 160
-fast = true
-skip-string-normalization = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "Django>=4.0,<5.0",
     "pre-commit",
 ]
+requires-python = ">=3.12"
 
 # Do not uncomment the line below. We need to be able to override the version via a file, and this
 # causes the "version" key in setup.cfg to be ignored.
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -2,11 +2,17 @@
...
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "metrics-utility"
+version = "0.1.0"
+description = "A metrics utility for Ansible"
+dependencies = [
+    "ruff",
+    "pytest",
+    "Django>=4.0,<5.0",
```
dhaustein, line 12: Question: Did we establish that the metrics-utility won't work with Django ver not higher than 4.2.x?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
```
dhaustein: Suggestion: Should probably lead to the official URL Astral mentions: `https://github.com/astral-sh/ruff-pre-commit`

Charlie Marsh's URL works but it will merely redirect to the official one, might as well use that directly.

dhaustein (changes requested):
I failed to make it work on my end. I can do `$ pre-commit run` manually and that will pickup the errors I made in my `test.py` file.

```bash
$ pre-commit run
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
```

But when I do `git add test.py` and `git commit -m "commit bad file"` it will happily let me.

```bash
$ vim test.py  # added the example from here https://realpython.com/ruff-python/#checking-for-errors

$ git add .

$ git commit -m 'test: test pre-commit hook'
[pr/42 59ac220] test: test pre-commit hook
 1 file changed, 18 insertions(+)
 create mode 100644 test.py

$ git status
On branch pr/42
nothing to commit, working tree clean
```

---

:fireworks:  **EDIT:** I had to re-run `$ pre-commit install` and then everything started working as expected. I suggest adding this as an extra install step to the Set-up steps.

dhaustein (approved):

pyproject.toml
himdel: awx is on 4.2.16 now

dhaustein: ah of course, that's the reason, I recall we talked about the need to align on _something_ with awx but I just couldn't remember what it was

himdel:
The changes look good :+1: ,

just, did you run `uv sync` after the last change? Looks like when I `git pr 42; uv sync`, it updates the lockfile, adding pre-commit, so maybe it wasn't up to date?

(Also, it warns about a missing requires-python, adding a `requires-python = ">=3.12"` to the `project` section in `pyproject.toml`, and a doing an `echo 3.12 > .python-version` seems to fix this .. unless we're keeping the version out deliberately?)

himdel:
In this PR description's Set-up section:

> in /metrics-utility root dir, create and activate your venv with python -m venv venv; source venv/bin/activate
install the deps using pip install

This bit seems surprising? uv manages virtual environmeents & dependencies for you, so I'd expect these to be something like "run `uv sync`" & `uv run ruff` / `uvx ruff` etc.? Or `uv venv ; source .venv/bin/activate
` instead of directly talking to virtualenv & pip?

himdel (approved):
LGTM :+1: , no warnings, no diff after sync, works for me :)

PR #43, opened 2025-01-20
Add testing data to gitignore

MilanPospisil:
[AAP-38383]

## Description

Adds testing files to gitignore - all files inside test folder will be part of the repo, excluding csv and json files, because those are part of *tar.gz files - if you want to extract them and look whats inside for debug purposes, you now dont have to delete them, they will not be included in repo.

Also, uninclude generated reports.

Folder shipped_data is unincluded, that should be used only for runtime debugging - any data that should not be part of the repo should live there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ metrics_utility/test/test_data/data/2024/**/*.csv
 
 # Reports
 /metrics_utility/test/test_data/reports/
+
+# Shipped data - for regular usage during developement
+/shipped_data
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,13 @@ testem.log
 
 # PyCharm
 .idea/*
+
+# Testing data
+metrics_utility/test/test_data/data/**
+!metrics_utility/test/test_data/data/2024/
+!metrics_utility/test/test_data/data/2024/**
+metrics_utility/test/test_data/data/2024/**/*.json
+metrics_utility/test/test_data/data/2024/**/*.csv
+
+# Reports
+/metrics_utility/test/test_data/reports/
--- a/.gitignore
+++ b/.gitignore
@@ -32,14 +32,13 @@ testem.log
 .idea/*
 
 # Testing data
-metrics_utility/test/test_data/data/**
-!metrics_utility/test/test_data/data/2024/
-!metrics_utility/test/test_data/data/2024/**
-metrics_utility/test/test_data/data/2024/**/*.json
-metrics_utility/test/test_data/data/2024/**/*.csv
+metrics_utility/test/test_data/data/**/*.csv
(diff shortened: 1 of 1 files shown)


msmagnanijr (approved):

art-tapin:
Could we clarify why we're not including all files in test_data under .gitignore? Would it make sense to add a broader rule like `metrics_utility/test/test_data/**` to ensure all test data is ignored? This might simplify maintenance and reduce potential clutter in the repo. Is there a specific reason for selectively ignoring only certain paths and file types?

himdel:
If the idea is to only keep the tar gz files, what about just

```
metrics_utility/test/test_data/data/**
-metrics_utility/test/test_data/data/**/*.tar.gz
```

(+ reports)? Or is 2024 important? Any new test data would still be a tar gz, but no longer 2024, right?

MilanPospisil:
> If the idea is to only keep the tar gz files, what about just
> 
> ```
> metrics_utility/test/test_data/data/**
> -metrics_utility/test/test_data/data/**/*.tar.gz
> ```
> 
> ? Or is 2024 important? Any new test data would still be a tar gz, but no longer 2024, right?

Well, gitignore is quite weird and rule of keeping only .tar.gz under specific folders (2024) was not easy to write, this is the only what worked :-)

But we can keep track of all .tar.gz, only if you gather new data for some purposes and dont want to include them to git, it would be harder to maintain.

On the other hand, we wants new data in git eventualy, so thats maybe simplier solution.

MilanPospisil:
> Could we clarify why we're not including all files in test_data under .gitignore? Would it make sense to add a broader rule like `metrics_utility/test/test_data/**` to ensure all test data is ignored? This might simplify maintenance and reduce potential clutter in the repo. Is there a specific reason for selectively ignoring only certain paths and file types?

Because input data for utility are *tar.gz files, but if you want to see what is inside them, you need to unpack them, but you dont want to add unpacked files to the git.

dhaustein:
Would it make sense to separate input and output data/folders in the future? That way we can keep anything we want as test data in git in the input folder(s) and ignore everything in the output folder so it's never commit to git even by accident?

MilanPospisil:
> Would it make sense to separate input and output data/folders in the future? That way we can keep anything we want as test data in git in the input folder(s) and ignore everything in the output folder so it's never commit to git even by accident?

Yes, working on that. I see that in readme, there is shipping_data folder, that should be used for normal work, for tests, we should use the test folder.

MilanPospisil:
Latest update - include everything in testing folder, except extracted csv and json files. Also uninclude from repo the shipped_data folder, that should be used for runtime purposes, for data that should not be part of repo.

himdel (approved):

PR #45, opened 2025-01-21
Add imported dependencies

himdel:
Issue: AAP-38424 / AAP-39220

Add all packages we import from as dependencies, using the same version that awx [currently uses](https://github.com/ansible/awx/blob/f57a9863d686505aa68b857c537d77e1b7cf538b/requirements/requirements.txt).
This is a step towards being able to run the utility in standalone environment for testing.

Adds:

    "boto3==1.35.96",
    "botocore==1.35.96",
    "distro==1.9.0",
    "insights-analytics-collector==0.3.2",
    "openpyxl==3.1.2",
    "pandas==2.2.1",
    "requests==2.32.3",
    "setuptools==70.3.0",

(Pretty sure most of these can be updated to latest, but starting from a known good state before we have tests.)

Changes:

 * Django from `>4<5` to `"django==4.2.16"` (because awx)
 * pre-commit, pytest & ruff from versionless to (no awx-related limitations)
   * `"pre-commit>=4.0.1",`
   * `"pytest>=8.3.4",`
   * `"ruff>=0.9.2",`
 * minimum python version to 3.11

This currently does nothing, except install successfully,
but the follow-up PR can focus on dealing with just imports from awx itself.


dhaustein (approved):

art-tapin (approved):
LGTM! 👍

The changes provide a solid foundation for making the metrics-utility more standalone by properly declaring all dependencies with specific versions. I've tested the branch and can confirm it maintains the same behavior as devel, with the expected warnings that will be addressed in follow-up work.

Special thanks to @himdel  for the quick support with troubleshooting the local installation issues.

Ready to merge! 🚀

PR #47, opened 2025-01-23
fix: switch to advisory_lock from DAB instead of pglock

dhaustein:
Hotfix for errors
```
ModuleNotFoundError: No module named 'awx.main.utils.pglock'
```
Caused by awx removing `django-pglocks` and switching to using `advisory_lock` from DAB in https://github.com/ansible/awx/commit/f57a9863d686505aa68b857c537d77e1b7cf538b

### DIFF
--- a/metrics_utility/automation_controller_billing/collector.py
+++ b/metrics_utility/automation_controller_billing/collector.py
@@ -14,7 +14,7 @@ from django.core.serializers.json import DjangoJSONEncoder
 from metrics_utility.automation_controller_billing.package.factory import Factory as PackageFactory
 
 from awx.main.utils import datetime_hook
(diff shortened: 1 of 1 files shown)


himdel:
LGTM, fixes a problem introduced by awx removing one of our dependencies :+1: 

Cc @Ladas 

While we still depend on the awx environment, this change could break on 2.4 (unless all 2.4 versions are using a recent enough django ansible base). But not doing it breaks on newer 2.5 so...

dhaustein:
Switching PR to draft. I need to address the fact that in 2.4 there is no DAB.

dhaustein:
Closed in favor of https://github.com/ansible/metrics-utility/pull/51 It will have the same change we wanted to do here

PR #49, opened 2025-01-23
configure linters to use uv+ruff and remove superlinter

ShaiahWren:
[[AAP-39264](https://issues.redhat.com/browse/AAP-39264)]

Remove the currently configured 'superlinter' and replace with ruff+uv, according to team consensus. This is an effort to 'check the check' in case the pre-commit hook fails. 

Note: After https://github.com/ansible/metrics-utility/pull/50 is merged the new linter should pass (pr50 runs ruff on the whole repo).

### Code Quality

## Notes for Reviewers


.github/workflows/main.yml
```diff
@@ -17,27 +17,38 @@ jobs:
...
+        with:
+          python-version-file: "pyproject.toml"
+
+      - name: Install the project with uv
+        run: uv sync --all-extras --dev
+
+      - name: Install Ruff
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff
```
dhaustein: suggestion: I don't think you need to install `ruff` again. Now that it's part of the project dependencies https://github.com/ansible/metrics-utility/blob/devel/pyproject.toml#L10 it should be installed once you invoke `uv sync` automatically. (Unless I misunderstood what uv is doing on that step...)

.github/workflows/main.yml
```diff
@@ -17,27 +17,38 @@ jobs:
...
+
+      # Update output format to enable automatic inline annotations.
+      - name: Run Ruff
+        run: ruff check --output-format=github .
+
+      - name: Configure JSCPD Linter
```
dhaustein: note: This is probably up for a team discussion, but I think sonar also tries to detect code duplication. We might not need jscpd anymore.

dhaustein (approved):

.github/workflows/main.yml
ShaiahWren: Part of SonarCloud's [exhaustive code analysis](https://medium.com/@rahmaadinda/sonarcloud-a-comprehensive-overview-of-static-analysis-features-d39884df1cdd) includes duplicate code detection.

.github/workflows/main.yml
ShaiahWren: That's what I was hoping too with `uv sync` but the ci didn't find ruff so I added the install. Could be that uv sync isn't doing anything here

himdel: I think this is not needed if you do `uvx ruff check...` in the next step.

(Or `uv run ruff check....`, or `uv tool install ruff` and then just `ruff check....`)

art-tapin (changes requested):
Hi! I ran into an issue while trying to set up metrics-utility from scratch using `uv sync`. I discovered that pandas 2.2.1 has compatibility issues with Python 3.13, specifically during the compilation of its Cython extensions. Here's a snippet of the error:

```
pandas/_libs/window/aggregations.cpython-313-x86_64-linux-gnu.so.p/pandas/_libs/window/aggregations.pyx.cpp:422:31:
error: standard attributes in middle of decl-specifiers
  422 |         #define CYTHON_UNUSED [[maybe_unused]]
      |                               ^
```

The issue appears to be related to how C++ attribute syntax is handled in Python 3.13. Since pandas 2.2.1 hasn't been fully updated for Python 3.13 support yet, I propose we update the `requires-python` field in pyproject.toml to:

```toml
requires-python = ">=3.11, <3.13"
```

art-tapin:
Great job, once you update requires-python in pyproject.toml, this PR will have my :heavy_check_mark:. Thanks @ShaiahWren

himdel (approved):

art-tapin (approved):
:1st_place_medal: 

Great! Ready to merge :sunglasses:

PR #51, opened 2025-01-27
AAP-39220 : Run ccspv2 build_report outside awx

himdel:
Issue: AAP-39220
Follows #45 

Make it possible to run the CCSPv2 `build_report` phase outside of the AWX environment.

Previously, when importing from `awx` failed, the util would try `$AWX_PATH` or `/awx_devel`, and fail when that failed.
Now, it will warn, but fall back to a `mock_awx` implementation.

Adds django-ansible-base as a dependency for the standalone mode,
imports `advisory_lock` from `ansible-base` when present (standalone or 2.5), or from `awx` otherwise (2.4).
(~~Either~~ replaces #47, ~~or replace this bit with #47.~~)

Mocks:

* `awx.prepare_env` - sets up `DJANGO_SETTINGS_MODULE` to point at `mock_awx/settings`
* `awx.main.get_awx_version` - hardcoded to `24.6.123` for now
* `awx.main.get_awx_http_client_headers` - license info hardcoded to `UNLICENSED`
* `awx.main.datetime_hook` - small wrapper around django `parse_datetime`, copied
* (`settings` - empty module)

(The idea is to hardcode these for testing, and later replace with a param or an os.environ.get when we need to.)

This does not do anything about anything missing for the gather phase of ccspv2, or any other reports or backends - still unaddressed:

* `Job`
* `Setting`
* `access_registry`
* `get_license`
* (and gather, renewal, s3, crc...)

Testing:

```
cd metrics-utility
gh pr checkout 51 # OR git fetch upstream pull/51/head ; git checkout FETCH_HEAD

./run-ccsp2-build
# check metrics_utility/test/test_data/reports/2024/04/CCSPv2-2024-04.xlsx
```

(The `run-ccsp2-build` script is just a convenience wrapper .. maybe we should move it to a `tools` dir or some such, ideas?)

(Possibly there should be an explicit param to allow running with the mocked env if we wanted to preserve 100% error compatibility too - but that means parsing the options before calling into ManagementUtility which deals with options and doesn't really align with the end goal?)

(Mocking awx is to help unblock testing, adding relative import dirs may not work so well for RPM installs.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = [
     "botocore==1.35.96",
     "distro==1.9.0",
     "django==4.2.16",
+    "django-ansible-base>=2025.1.3",
     "insights-analytics-collector==0.3.2",
     "openpyxl==3.1.2",
     "pandas==2.2.1",
--- a/metrics_utility/__init__.py
+++ b/metrics_utility/__init__.py
@@ -15,8 +15,8 @@ def manage():
         sys.path.append(awx_path)
         spec = importlib.util.find_spec('awx')
         if spec is None:
-            sys.stderr.write(f"Automation Controller modules not found in {awx_path}\n")
-            exit(1)
+            sys.stderr.write(f"Automation Controller modules not found in {awx_path} (AWX_PATH)\n")
+            sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__), '..', 'mock_awx')))
 
     import django
     from awx import prepare_env
--- a/metrics_utility/__init__.py
+++ b/metrics_utility/__init__.py
@@ -15,7 +15,7 @@ def manage():
         sys.path.append(awx_path)
         spec = importlib.util.find_spec('awx')
         if spec is None:
-            sys.stderr.write(f"Automation Controller modules not found in {awx_path} (AWX_PATH)\n")
+            sys.stderr.write(f"Automation Controller modules not found in {awx_path} (AWX_PATH). Using mock and continuing.\n")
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "botocore==1.35.96",
     "distro==1.9.0",
     "django==4.2.16",
+    "django-ansible-base>=2025.1.3",
```
Ladas, line 14: will this go to setup.cfg used for prod dependencies?

Assuming this dependency won't be present in 2.4, is there a way to do optional dependency?

Ladas: seems like there is extras_require https://stackoverflow.com/questions/10572603/specifying-optional-dependencies-in-pypi-python-setup-py 

so we should be able to use that in prod mode, so it fetches dab only when it's there?

Ladas (approved):
👍 great step forward

pyproject.toml
himdel: My understanding is that the RPM build would only use setup.cfg, but not pyproject.toml.

If it uses both, we'd probably need to move the standalone util into a separate repo..

(Asking Satoe to make sure...)

himdel: OK, update from Satoe .. the RPM builds *might* use `pyproject.toml` first if both exist, but it is possible to explicitly say to use `setup.cfg` only for these.

So.. probably something to keep in mind when doing a release, but .. ok for now?

dhaustein (approved):

MilanPospisil (approved):
LGTM :+1:

PR #52, opened 2025-01-29
Run ccsp2 gather_automation_controller_billing_data outside awx

himdel:
Issue: AAP-40391
Follows #51
Closes #48 
Closes #65 

Make it possible to run the CCSPv2 gather\* phase and the renewal report outside of the AWX environment.
The groundwork was done in #51, this mostly deals with setting up the database and adding some missing functions and settings.

`run-ccsp2-gather` actually works now (on an empty db), `run-renewal` also works now
adds `psycopg`, more mock settings
adds a sample awx db schema (changes from #48)

enables debug logging in standalone (changes from #65)
enables `USE_TZ` for django 4 dates to work (and undoes a bit from #66 switching to timezone)

(and runs `uvx ruff format` on the new files)

---

How the db works is that we run the [`cloudservices/postgres`](https://quay.io/cloudservices/postgres) container,
which can load arbitrary sql files on init - so we init the awx role, and import an awx schema,
and mock_awx settings point to that db, which means gather now talks to that db and can have arbitrary mocked data there.


run-ccsp2-gather
```diff
@@ -0,0 +1,52 @@
...
+export METRICS_UTILITY_REPORT_END_USER_COMPANY_NAME="Customer A"
+export METRICS_UTILITY_REPORT_END_USER_COUNTRY="US"
+export METRICS_UTILITY_REPORT_END_USER_STATE="TX"
+export METRICS_UTILITY_REPORT_H1_HEADING="CCSP NA Direct Reporting Template"
+export METRICS_UTILITY_REPORT_PO_NUMBER="123"
+export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
+export METRICS_UTILITY_REPORT_SKU="MCT3752MO"
+export METRICS_UTILITY_REPORT_SKU_DESCRIPTION="EX: Red Hat Ansible Automation Platform, Full Support (1 Managed Node, Dedicated, Monthly)"
+
+SCHEMA=`realpath "metrics_utility/test/test_data/schemas/awx/latest.sql"`
```
MilanPospisil, line 22: Maybe this part of running, migrating and connecting to postges can be in shared script?

himdel: No point, this is just a helper script to show how to run this.

But eventually this will be shared a CI workflow.
(Or a compose file if we go with the pattern in #57)

mock_awx/awx/conf/license/__init__.py
```diff
@@ -0,0 +1,12 @@
+class Unlicense(object):
+    def validate(self):
+        return dict(
+            license_type='UNLICENSED',
+            product_name='Red Hat Ansible Automation Platform',
```
Ladas: nit: just state AWX as product name (similar below)

Ladas (approved):
👍

PR #56, opened 2025-01-31
Milan snapshot test

MilanPospisil:
[AAP-39142]

### How the test works

This PR has dozens of files changed, but only 3 are relevant - rest are generated testing data.

This test aims to generate many xlsx files that can be used as snapshots for the further testing.
Idea is that:

1) Many xlsx files will be in the repo along with stored command arguments and env variables that were used for generating the xlsx files. 

2) Test will run and use the same params and env variables as original xlsx file (read from its respective json file), run command, generate new xlsx files and compare them to old ones. The comparsion is done by textual content inside spreadsheet, not by some binary exact coparsion, which would not work, since each xlsx file contains its own timestamp and possible other different metadata line unique GUIDs.

3) PR also extending tests (that was not in the issue description) for non snapshot variants that is comparing  newly generated xlsx files of two different commands that should generate the same spreadsheet, like command pairs:

build_report --months=2024-02
build_report --since=2024-02-01 --until=2024-02-29

The range spans through the whole month, so it should generate the same spreadsheet.

Note that spreeadsheet comparsion compares only texts inside particular cells, but there are exceptions like created field, that contains date and this will differ and date range, which is either particular month or date range. Those 2 cells are ignored in snapshot test.

### Why is the test important

This test will be useful in developement and mainly in refactoring, so we will ensure that code is still producing the same results as it was before. When introducing some changes like new feature or bug fix, the snapshot test will fail. We can then see what failed and if the failure was expected - for example we are repairing / extending some data part and if only that data part is changed, thats ok, if it changed also something unexpected, we may alter something unwanted by mistake. But for every intentional change that alter command workings, we will have to generate the historical test files again.

### Description of source code files

Now description of each source code file:

[metrics_utility/test/snapshot_tests/CCSP/CCSP_snapshot_generator.py](https://github.com/ansible/metrics-utility/pull/56/files#diff-94bbcebea345cd40f3384be1bd5c600a7990dc61fdfad9ea8c652209d6ab7d76)

Run this script to generate reports for further tests:
python -m metrics_utility.test.snapshot_tests.CCSP.CCSP_snapshot_generator

This is generating reports with different parameters and env_variables. It generates CCSPv2 and also CCSP reports.
For each report, it first generates <file_name>.json file with the settings, that can look like this:

{
    "env_vars": {
        "METRICS_UTILITY_SHIP_TARGET": "directory",
        "METRICS_UTILITY_SHIP_PATH": "/awx_devel/awx-dev/metrics-utility/metrics_utility/test/test_data",
        "METRICS_UTILITY_PRICE_PER_NODE": "15.01",
        "METRICS_UTILITY_REPORT_SKU": "CSP8794CE",
        "METRICS_UTILITY_REPORT_SKU_DESCRIPTION": "EX: Red Hat OpenShift Container Platform, Premium Support (1 Node, Monthly)",
        "METRICS_UTILITY_REPORT_H1_HEADING": "CCSP EU Direct Reporting Template",
        "METRICS_UTILITY_REPORT_COMPANY_NAME": "Partner B",
        "METRICS_UTILITY_REPORT_EMAIL": "support@partner.com",
        "METRICS_UTILITY_REPORT_RHN_LOGIN": "admin_user",
        "METRICS_UTILITY_REPORT_PO_NUMBER": "456",
        "METRICS_UTILITY_REPORT_END_USER_COMPANY_NAME": "Customer B",
        "METRICS_UTILITY_REPORT_END_USER_CITY": "Rivertown",
        "METRICS_UTILITY_REPORT_END_USER_STATE": "CA",
        "METRICS_UTILITY_REPORT_END_USER_COUNTRY": "Canada",
        "METRICS_UTILITY_REPORT_TYPE": "CCSPv2"
    },
    "params": [
        "manage.py",
        "build_report",
        "--since=2024-02-05",
        "--until=2024-04-30",
        "--force"
    ],
    "custom_params": {
        "run_command": "Yes",
        "generated": "2025-01-31"
    }
}

And there is folder with the same name as json file that contains generated report that has been generated using subprocess call using this env variables and input command params.

For example:
snapshot_def_2024-02.json
snapshot_def_2024-02/result.xlsx

While env_vars are used for setting env vars for subprocess and params are direct command array of params, custom_params can contains additional paramters for tests:

1) "run_command" : if Yes, it will run command as subprocess, in future, we may add also ability to run command as direct function call (without suprocess), so we can mock datetime.now to test also relative time parameters like --since=5months, which will differ based on current date.

2)"generated" - date when it was generated if we will need it

Generator will generate many of those files with different ranges and different env_variables (except report type - which is fixed to either CCSPv2 or CCSP, ship path and ship target, which is always the same).

[metrics_utility/test/snapshot_tests/CCSP/CCSP_snapshot_test.py](https://github.com/ansible/metrics-utility/pull/56/files#diff-c1fa9d89515a1c4ca0c58682b36ca751c8e63a70d3b24931783d886861b5b259)

This is scanning all *.json files of test definitions for each definition:
1) Finds xlsx report that was generated in the past by generator
2) generate new xlsx file using the same definition found in the *.json
3) Compare those files, if they are not equal, fail the test

Additionaly, it runs comparsion for month 2024-02 and 2024-03 like:
build_report --months=2024-02
build_report --since=2024-02-01 --until=2024-02-29

Both reports are compared if they are equal.

[metrics_utility/test/snapshot_tests/snapshot_utils.py](https://github.com/ansible/metrics-utility/pull/56/files#diff-22b5b674a5432e87b324b97ea1dd14fa21aad140e962f014aaaa2fc347477ece)

This file contains support utilities for the files above. Many of those funtions can be reused for later snapshots tests for another command.

### Ignoring resource warning failures

openpyxl has some troubles with closing the files, even the xlsx files are closed, the pytest throws error about unclosed files.

However, adding marker:
@pytest.mark.filterwarnings('ignore::ResourceWarning')

for the tests that are using this library solves the issue.

### Steps to Test

This runs generator (you must be at root folder of metrics utility project):

python -m metrics_utility.test.snapshot_tests.CCSP.CCSP_snapshot_generator

Once testing files are generated, run particular test:

SETUPTOOLS_USE_DISTUTILS=true pytest -s

### Expected Results

The tests should pass.

If you want to test its failure, open some xlsx file, for example:
/metrics-utility/metrics_utility/test/snapshot_tests/CCSP/data/CCSPv2/snapshot_def_2024-02/report.xlsx

Enter in some random empty cell text Error

Then you should see assertion error that empty cell does not equals to Error. 

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers


metrics_utility/automation_controller_billing/collector.py
```diff
@@ -1,10 +1,18 @@
 import contextlib
 import json
 import logging
+import importlib.util
 
 import insights_analytics_collector as base
 from awx.main.utils import datetime_hook
-from awx.main.utils.pglock import advisory_lock
+
+if importlib.util.find_spec('ansible_base') is None:
```
MilanPospisil: This is from Martins unfinished PR:
if importlib.util.find_spec('ansible_base') is None:
    # 2.4
    from awx.main.utils.pglock import advisory_lock
else:
    # 2.5
    from ansible_base.lib.utils.db import advisory_lock
    
    Otherwise, this PR would be blocked.

metrics_utility/test/snapshot_tests/snapshot_utils.py
```diff
@@ -0,0 +1,245 @@
...
+    print(f'Opening {original_report_path}')
+    o_wb = openpyxl.load_workbook(filename=original_report_path)
+
+    try:
+        compare_worksheets(g_wb, o_wb, 0, ['H1', 'B5'])
+        compare_worksheets(g_wb, o_wb, 1, [])
+        compare_worksheets(g_wb, o_wb, 2, [])
+    finally:
+        g_wb.close()
+        o_wb.close()
```
dhaustein, line 262: Note: A custom context manager that will close the file automatically might be wort exploring in the future.

metrics_utility/test/snapshot_tests/snapshot_utils.py
```diff
@@ -0,0 +1,245 @@
...
+
+    try:
+        compare_worksheets(g_wb, o_wb, 0, ['B5'])
+        compare_worksheets(g_wb, o_wb, 1, [])
+    finally:
+        g_wb.close()
+        o_wb.close()
+
+
+def compare_worksheets(workbook_generated, workbook_original, sheet_number, exceptions: List[str]):
```
dhaustein: Suggestion (non-blocking): Might be interesting to try comparing excel sheets with pandas. I briefly checked how that is done and it should result in cleaner code. However, that would probably require to switch away from openpyxl altogether.

metrics_utility/test/snapshot_tests/snapshot_utils.py
```diff
@@ -0,0 +1,245 @@
...
+def get_param_value(params: List[str], name):
+    for param in params:
+        if param.startswith(name + '='):
+            key, value = param.split('=')
+            if key == name:
+                return value
+    return None
+
+
+def get_entry_point_directory():
```
dhaustein: Nitpick: The functions in this file are all missing a docstring. Since these are utility functions, other people might want to use them in the future too and would be looking for documentation on how to do that.

metrics_utility/test/snapshot_tests/snapshot_utils.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def create_directory_if_not_exists(directory_path):
+    try:
+        os.makedirs(directory_path, exist_ok=True)
+    except Exception as e:
+        print(f'Error creating directory: {e}')
+
+
+def parse_json_file(file_path):
```
dhaustein: Suggestion: I wonder if all this loading/reading a JSON file can be skipped by having the input values in a python files directly. I don't think there is a specific need to have the snapshot env vars, params etc. in a JSON file so you might save yourself the complexity of de-serializing them.

MilanPospisil: Maybe, I wanted those files to be clearly associated with the xlsx files and tests can then run quite independently on whats in the generator. But we can disscuss it and possibly change it in follow up PR?

dhaustein: Sure thing, this is not a blocking suggestion

Ladas (approved):
👍 looks great

ShaiahWren (approved):
Looks good. Needs rebase

dhaustein (approved):

PR #55, opened 2025-01-30
AAP-39214 - Fine tune pytest

ShaiahWren:
[AAP-39214]

## Description

'Fine-tuning' - making the CCSPv2 test more robust by checking all column values. 

## Testing

See `/metrics-utility/metrics_utility/test/ccspv_reports/test_CCSPv2.py`

### Prerequisites

None

### Steps to Test

Before running test, enter awx container:

```
 docker exec -it tools_awx_1 /bin/sh -c "
    cd awx-dev/metrics-utility &&
    . /var/lib/awx/venv/awx/bin/activate &&
    exec /bin/sh
 "
```

From /metrics-utility run `SETUPTOOLS_USE_DISTUTILS=true pytest -s`

### Expected Results

All tests should pass

## Required Actions

- [ ] Requires documentation updates

- [ ] Requires downstream repository changes

- [ ] Requires infrastructure/deployment changes

- [ ] Requires coordination with other teams

- [ ] Blocked by PR/MR: #XXX

### Code Quality

- [x] Code is properly linted and formatted.
- [x] All tests (existing and new) pass successfully.
- [x] Tests are added or updated as needed.
- [x] Code includes relevant comments for complex sections.
- [x] Changes are reviewed and approved by at least two team members.
- [ ] Documentation is updated where applicable.  
       - If updates are required in the [handbook](https://github.com/ansible/handbook), create a separate PR for the handbook repo.

## Notes for Reviewers

Add any additional context or instructions for reviewers here - for example screenshots if needed.

PR #60, opened 2025-02-04
test  simple ccsp file

ShaiahWren:
[AAP-39222]

## Description

Similar to the [CCSPv2](https://github.com/ansible/metrics-utility/pull/55) test where was checked all columns and values from a built CCSPv2 report, the work at hand builds and tests the more simple CCSP one.

Much of the logic is the same with a few key changes:
- env vars now exclude 'END_USER' data such as city, state, country, and so forth.
- report excludes the 'Usage by Organizations' sheet.
- to more easily test the column in 'Usage Reporting', we will start on line 14 to grab the column header and  then traverse from the top. To start at the very top introduces too much confusion because of how many blank lines there are in the top row.
- added conftest.py file to store fixtures and static methods. Refactored as needed. This change ensures that the check for SonarCloud duplication passes. Without the shared helper functions, codebase dup is > 13%. Yikes!
### Prerequisites

Enter the awx virtual env

`docker exec -it tools_awx_1 /bin/sh -c "
    cd awx-dev/metrics-utility &&
    . /var/lib/awx/venv/awx/bin/activate &&
    exec /bin/sh
 "`

### Steps to Test

1. From /metrics-utility run `SETUPTOOLS_USE_DISTUTILS=true pytest -s`

### Expected Results

Both tests should pass

## Required Actions

- [x] Requires follow-on pr once we are running in standalone mode
- Add a separate fixture dir and load them from conftest.py 
- Manage wb close with context manager 

### Code Quality

## Notes for Reviewers

### DIFF
--- a/metrics_utility/test/ccspv_reports/conftest.py
+++ b/metrics_utility/test/ccspv_reports/conftest.py
@@ -4,46 +4,51 @@ import openpyxl
 
 class Helpers:
     @staticmethod
-    def validate_sheet_tab_names(file_path, EXPECTED_SHEETS):
+    def validate_sheet_tab_names(file_path, expected_sheets):
         """Test the sheet names in the Excel file."""
 
         wb = openpyxl.load_workbook(file_path)
         try:
             actual_tab_names = wb.sheetnames
             assert actual_tab_names == list(
-                EXPECTED_SHEETS.keys()
+                expected_sheets.keys()
             ), "Sheet names do not match."
         finally:
             wb.close()
 
     @staticmethod
-    def validate_sheet_columns(file_path, EXPECTED_SHEETS, usage_reporting_min_row):
+    def normalize_column(col):
+        if not col:
+            return ""
+        return col.strip().replace("\n", " ").lower()
+
+    @staticmethod
+    def validate_sheet_columns(file_path, expected_sheets, usage_reporting_min_row):
         """Test the column names for each sheet."""
 
-        def normalize_column(col):
-            return col.strip().replace("\n", " ").lower() if col else ""
+        # For the 'Usage Reporting' sheet, start at specific row with value for ease of traversing
+        def get_min_row(sheet_name):
+            return usage_reporting_min_row if sheet_name == "Usage Reporting" else 1
+
+        # Function returns column headers
+        def get_column_headers(expected_column_data):
+            expected_column_headers = []
+            for column_group in expected_column_data:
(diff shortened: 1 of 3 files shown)


metrics_utility/test/ccspv_reports/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+                            cell.value for row in sheet.iter_rows(min_row=2, min_col=col_index, max_col=col_index)
+                            for cell in row
+                        ]
+
+                        # Assert column values
+                        assert actual_column_values == expected_column_values, (
+                            f"Column values do not match for column '{expected_col_name}' in sheet '{sheet_name}'"
+                        )
+        finally:
+            wb.close()
```
dhaustein: Suggestion: I think a context manager that open/closes the file could be a cleaner solution.

metrics_utility/test/ccspv_reports/conftest.py
```diff
@@ -0,0 +1,91 @@
+import os
+import pytest
+import openpyxl
+
+class Helpers:
```
dhaustein: Nitpick: Outright making a class is probably overkill, especially if you fill it with only static methods. Might as well just create three functions. Unless the plan is to bundle state with them later on ofc.

metrics_utility/test/ccspv_reports/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+    @staticmethod
+    def validate_sheet_tab_names(file_path, expected_sheets):
```
dhaustein: Suggestion: Please add docstrings to the methods in this file.

dhaustein (approved):
Left a few nitpicks otherwise LGTM.

I have one suggestion for the future however. While having the fixtures in `conftest.py` is absolutely fine, it does not scale well when the number of fixtures increases.

A trick I picked up in EDA was to create a number of semantically distinct "plugin" files for fixtures and load them in conftest at runtime. https://github.com/ansible/eda-qa/blob/main/conftest.py#L13-L39

But this is not a blocker or even a big deal right now, but might be something to think about for the future. :slightly_smiling_face:

MilanPospisil (approved):
Tested and works :-)
LGTM :+1:

metrics_utility/test/ccspv_reports/conftest.py
ShaiahWren: Good idea, why don't I implement this in a follow-on pr once we are running in standalone mode.

PR #65, opened 2025-02-11
Fix bug in standalone mode - datetime utc and logger

MilanPospisil:
[AAP-40143]

## Description

Two changes:

1) used standard datetime timezones instead of django ones. This fix error:

Traceback (most recent call last):
  File "/home/milan/ansible_galaxy/awx/awx-dev/metrics-utility/manage.py", line 6, in <module>
    manage()
  File "/home/milan/ansible_galaxy/awx/awx-dev/metrics-utility/metrics_utility/__init__.py", line 28, in manage
    utility.execute()
  File "/home/milan/ansible_galaxy/awx/awx-dev/metrics-utility/metrics_utility/management_utility.py", line 58, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/milan/.local/lib/python3.11/site-packages/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/milan/.local/lib/python3.11/site-packages/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/milan/ansible_galaxy/awx/awx-dev/metrics-utility/metrics_utility/management/commands/build_report.py", line 79, in handle
    opt_since = parse_date_param(opt_since)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/milan/ansible_galaxy/awx/awx-dev/metrics-utility/metrics_utility/automation_controller_billing/helpers.py", line 31, in parse_date_param
    parsed_date = parsed_date.replace(tzinfo=timezone.utc)
                                             ^^^^^^^^^^^^
AttributeError: module 'django.utils.timezone' has no attribute 'utc'

2) Repaired logger - PR introduces new env variable "STANDALONE_MODE", this is set when awx is not found. Then if this env var is set, logger is set differently. Previous logger is set to work inside awx, but this setting did not worked in standalone.

Also handlers are added only when dont exist, this prevents duplicating handers, which can happen when __init__ is loaded multiple times, this can happen in tests.

## Testing
Described in issue.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

### DIFF
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -68,8 +68,7 @@ class Command(BaseCommand):
                 handler.setFormatter(logging.Formatter('%(message)s'))
                 self.logger.addHandler(handler)
                 self.logger.propagate = False
-
-        if os.getenv('STANDALONE_MODE') is not None:
+        else:
             self.logger.setLevel(logging.DEBUG)  # Ensure the logger captures all messages
             if not self.logger.handlers:  # If no handlers exist, add one
                 standalone_handler = logging.StreamHandler()
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -76,7 +76,7 @@ class Command(BaseCommand):
                 standalone_handler.setLevel(logging.DEBUG)
                 standalone_handler.setFormatter(logging.Formatter('%(message)s'))
                 self.logger.addHandler(standalone_handler)
-            
+
             self.logger.propagate = False  # Ensure logs do not propagate to other handlers
 
     def handle(self, *args, **options):
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -5,7 +5,7 @@ import os
 from dateutil.parser import parse as date_parse
 from dateutil.relativedelta import relativedelta
 from django.core.management.base import BaseCommand
-from django.utils import timezone
+from datetime import timezone
 
 from metrics_utility.automation_controller_billing.dataframe_engine.factory import \
     Factory as DataframeEngineFactory
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -58,11 +58,26 @@ class Command(BaseCommand):
 
     def init_logging(self):
         self.logger = logging.getLogger('awx.main.analytics')
-        handler = logging.StreamHandler()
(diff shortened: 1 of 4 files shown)


metrics_utility/management/commands/build_report.py
```diff
@@ -58,11 +58,26 @@ def add_arguments(self, parser):
...
+            # Prevent duplicate handlers
+            if not self.logger.handlers:
+                self.logger = logging.getLogger('awx.main.analytics')
+                handler = logging.StreamHandler()
+                handler.setLevel(logging.DEBUG)
+                handler.setFormatter(logging.Formatter('%(message)s'))
+                self.logger.addHandler(handler)
+                self.logger.propagate = False
+
+        if os.getenv('STANDALONE_MODE') is not None:
```
Ladas:
```
if os.getenv('STANDALONE_MODE') is None:
  ...
else:
  ...
```  
 
 is probably cleaner

dhaustein: You can do just
```python
if os.getenv('STANDALONE_MODE'):
  ...
else:
  ...
```
None in Python is Falsey

EDIT: Oh wait your comment was for the whole thing. :melting_face:

himdel:
Ah, django 4 defaults to `USE_TZ = False`, but awx defaults have `USE_TZ = True`.  So this is already fixed in https://github.com/ansible/metrics-utility/pull/52 .

But .. it should be safe to switch to native datetime too :+1: , just need to do it in all the places...

Missing 2:

```
metrics_utility/automation_controller_billing/collectors.py
9:from django.utils.timezone import now, timedelta

metrics_utility/management/commands/gather_automation_controller_billing_data.py
13:from django.utils import timezone
```

---

> `STANDALONE_MODE`

Why the name? It feels like if we're introducting an env variable, it should start with `METRICS_`.
But .. why introduce an env variable if it's only consumed by the same process? Why not  a regular variable? 

Also, shouldn't the place setting up everything for the standalone mode just set up loggers correctly too? So we don't need such logic elsewhere?

If `manage()` sets the loggers right, we wouldn't need any changes to `build_report` or any new env vars.

(It might make sense to split this into 2 PRs if you want to do both changes.)

MilanPospisil:
Closing, this changes will be addressed in https://github.com/ansible/metrics-utility/pull/52

PR #66, opened 2025-02-12
Add prints

MilanPospisil:
[AAP-39143]

## Description

Addition of some debug prints if --verbose flag is on.
debug_utils contains 3 functions:

1) One prints text if --verbose is on
2) Second prints dataframe with many excluced columns, so focus is to print important columns
3) Third is usable for debugging purposes in VS Code for easily setting params and env vars if necessary

jobhost summary and event has prints for better debugging.

### DIFF
--- a/debug_utils.py
+++ b/debug_utils.py
@@ -1,5 +1,6 @@
 from pprint import pprint
-import os, sys
+import os
+import sys
 
 def print_debug(text):
     if '--verbose' not in sys.argv:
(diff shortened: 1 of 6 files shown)


himdel (approved):
LGTM, doesn't affect anything without `--verbose`

(the `timezone` changes seem related to #65, I can address these in #52 too)

Ladas (approved):

PR #61, opened 2025-02-05
[AAP-38577] Docs: Refactor README.md & Relocate Legacy Installation Guide

art-tapin:
## **Description**  

  

- **What is being changed?**  
  - The existing `README.md` has been updated.  
  - The previous `README.md` has been moved to `/docs/old-readme.md` for reference.  
  - A few updates in installation instructions.  

- **Why is this change needed?**  
  - The outcome of the [Spike: Research and Define Documentation Structure & Storage](https://issues.redhat.com/browse/AAP-38577)
  - Recent AWX decoupling work has impacted the installation process, requiring adjustments to the setup documentation.  
  - The current documentation strategy mandates separating detailed setup guides into structured files under `/docs`.  
  - The `old-readme.md` still contains useful information and will serve as a reference for further documentation improvements by the developers contributing the changes.  

- **How does this change address the issue?**  
  - The new `README.md` includes relevant, working instructions and is more concise and easier to read.
  - The `/docs` directory will house additional in-depth guides, which will be created based on information from `old-readme.md`.  
  - Upcoming documentation additions *might* include:  
    - **Installation Guide**  
    - **Configuration Guide**  
    - **Storage Adapters Guide**  
    - **Report Types Guide**  

---

### **Steps to Test**  
- Running the Installation Steps in the provided documentation, ensuring the expected outcome.

### **Expected Results**  
- `uv pip install .` should complete successfully.  
- CCSPv2 report should be generated without errors.  
- Controller environment commands should function correctly.  

---

## **Required Actions**  

- [x] Requires documentation updates  
  - The old `README.md` is now `/docs/old-readme.md` and will be refined further.  
  - The following guides should be created based on `old-readme.md`:  
    - Installation Guide  
    - Configuration Guide  
    - Storage Adapters Guide  
    - Report Types Guide  
- [ ] Requires downstream repository changes - mentioning in the Ansible Handbook: https://github.com/ansible/handbook/pull/308

---

### **Code Quality**  

- [x] Documentation follows the agreed-upon structure.  
- [ ] All setup instructions are tested and verified.  
- [ ] Changes are reviewed and approved by at least two team members.  

---

## **Notes for Reviewers**  

- The separation of `old-readme.md` ensures we don’t lose valuable information while keeping the main `README.md` clean and structured.  
- Future documentation expansion will cross-reference details from the `old-readme.md`.  
- Let me know if additional refinements are required before merging.  
[...]

PR #69, opened 2025-02-17
Adding an example of 'date' in the README as it is required. Otherwise, it will fail

larrymou9:
If we don't add a date, we will end up the following error. 

`Traceback (most recent call last):
  File "/root/metrics-utility/manage.py", line 6, in <module>
    manage()
  File "/root/metrics-utility/metrics_utility/__init__.py", line 28, in manage
    utility.execute()
  File "/root/metrics-utility/metrics_utility/management_utility.py", line 58, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/lib/awx/venv/awx/lib64/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/metrics-utility/metrics_utility/management/commands/build_report.py", line 124, in handle
    extra_params=extra_params).create()
                               ^^^^^^^^
  File "/root/metrics-utility/metrics_utility/automation_controller_billing/dataframe_engine/factory.py", line 26, in create
    return (self._get_db_dataframe_host_metric_usage().build_dataframe(),)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/metrics-utility/metrics_utility/automation_controller_billing/dataframe_engine/db_dataframe_host_metric.py", line 24, in build_dataframe
    for data in self.extractor.iter_batches():
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ExtractorDirectory.iter_batches() missing 1 required positional argument: 'date'`

So I have added this in the readme as an example so that people can run it without these problems

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,14 @@ Recent changes in AWX have decoupled some dependencies, meaning certain componen
 4. Generate renewal guidance report:
 
    ```bash
-   export METRICS_UTILITY_REPORT_TYPE=RENEWAL_GUIDANCE
-   python manage.py build_report --since=12months --ephemeral=1month
+      # Set extra ENV VARs for report generation purposes
+      export METRICS_UTILITY_SHIP_TARGET=controller_db
+      export METRICS_UTILITY_REPORT_TYPE=RENEWAL_GUIDANCE
+      export METRICS_UTILITY_SHIP_PATH=/awx_devel/awx-dev/metrics-utility/shipped_data/billing
+
+      # Builds report covering 365days back by default
+      python manage.py build_report --since=12months --ephemeral=1month
+      # or metrics-utility build_report --since=12months --ephemeral=1month
    ```
 
 ## Documentation


larrymou9:
@art-tapin @Ladas I made this slight change as I was following the "new" README and I was not able to properly execute renewal guidance report. Please review!

himdel:
We might want to revert #61 (or, parts of it)

The new README is missing most of the variables we need, and calling a convenience script meant for dev testing only.

Also, `uv` is NOT how a customer would run it, it's just our dev env.

README.md
```diff
@@ -84,8 +84,14 @@ Recent changes in AWX have decoupled some dependencies, meaning certain componen
 4. Generate renewal guidance report:
 
    ```bash
-   export METRICS_UTILITY_REPORT_TYPE=RENEWAL_GUIDANCE
-   python manage.py build_report --since=12months --ephemeral=1month
+      # Set extra ENV VARs for report generation purposes
+      export METRICS_UTILITY_SHIP_TARGET=controller_db
```
Ladas, line 88: based on above steps, this example should be generating CCSP report, not renewal guidance?

(because renewal guidance doesn't need to call the collector step now)

PR #70, opened 2025-02-18
Add more complex ccsp tests

MilanPospisil:
[AAP-40225]

## Description

Test that assert correct values for the sheet generated by command for data in 2025-02-13.

Data contains events (playbooks, collections, roles, modules), also more organizations a duplicated hosts.
The test aims to assert more diverse data than tests before. And also, this is the first test that tests the events sheet.

## Testing

SETUPTOOLS_USE_DISTUTILS=true pytest -s ./metrics_utility/test/ccspv_reports/complex_test_CCSPv2.py

Test should pass


himdel (approved):

Ladas (approved):
Looks great 👍

PR #71, opened 2025-02-19
[AAP-38819] - Setting up pytest + S3 (Minio) + example test

himdel:
Issue: AAP-38819
Closes #57 

I was testing #57 after #52 by adding postgres to the mix,
and managed to get it working (I think :))

cc @msmagnanijr 
This is your #57 rebased on top of current devel,, with a commit adding postgres, switching to quay repositories, and working around permission issues with uv.

`gather_automation_controller_billing_data` runs successfully now :) 

---

### What is being changed?  
- **Minio setup** for local and CI testing.  
- Addition of an **initial test scenario** to validate `CCSPv2` report generation with **S3 backend**.  
- Integration with **pytest** to run automated tests using Minio as a simulated backend.  

### Why is this change needed?  
- Enables testing of `metrics-utility` S3 storage functionality without relying on an external environment.  
- Facilitates **test automation** in the CI pipeline.  
- Ensures that **report generation** works correctly with the S3 backend before full implementation.  

### How does this change address the issue?  
- Uses **Minio** as a local S3 backend for testing.  
- Defines the necessary setup to run locally and in CI.  
- Implements an **initial test case** covering `CCSPv2` report generation.

PR #57, opened 2025-01-31
WIP - [AAP-38819] - Setting up pytest + S3 (Minio) + example test

msmagnanijr:
# [AAP-38819] Setting up pytest + S3 (Minio) + example test  

[**AAP-38819**  ](https://issues.redhat.com/browse/AAP-38819)

## Description  

WIP

### What is being changed?  
- **Minio setup** for local and CI testing.  
- Addition of an **initial test scenario** to validate `CCSPv2` report generation with **S3 backend**.  
- Integration with **pytest** to run automated tests using Minio as a simulated backend.  

### Why is this change needed?  
- Enables testing of `metrics-utility` S3 storage functionality without relying on an external environment.  
- Facilitates **test automation** in the CI pipeline.  
- Ensures that **report generation** works correctly with the S3 backend before full implementation.  

### How does this change address the issue?  
- Uses **Minio** as a local S3 backend for testing.  
- Defines the necessary setup to run locally and in CI.  
- Implements an **initial test case** covering `CCSPv2` report generation.  

---

### Expected Results  

- ✅ The **Minio environment** starts successfully.  
- ✅ The **metrics-utility-test** bucket is created successfully.  
- ✅ **S3 report generation tests** run and pass without errors.

PR #72, opened 2025-02-20
Update github workflows - enable postgres, minio

himdel:
Issue: AAP-39227

Update github workflows - enable postgres & minio (almost enables pytest)
    
* main: rename to pytest, remove unused workflow inputs, ~~enable pytest~~
* update action versions to latest, remove uv version constraint
* consistent checkout & uv & python setup
* add ruff format check output (but don't fail yet), move ruff check to pr-checks
* change sonarcloud to pull_request_target - uses secrets
* update sonarcloud action to new name, pass token via env
* quote names, Literal Block Scalar for shells, Folded Block Scalar for args

And run awx postgres & minio before pytest so they're available during - logic from #57 / #71 , but without quay containers which would need logins.

And enable dependabot updates for github action versions.

The one thing this DOESN'T do yet is enable pytest - moving this to the next PR that makes sure exisiting tests can run both in github and with docker compose - https://github.com/ansible/metrics-utility/pull/74


Ladas (approved):
👍

PR #74, opened 2025-02-26
Pytest enable

himdel:
Issue: [AAP-39227](https://issues.redhat.com/browse/AAP-39227)
Follows #72

This enables pytest,
updates test paths to be relative to metrics utility root,
and updates snapshots to match these paths.
Also allows more characters in paths for the snapshot util - such as `./`
And renames complex_test_... to test_complex_... to make pytest find it.
And mentions how to run the tests in the README.

After this, tests can run the same way on CI & locally with docker compose.
Both `.github/workflows/pytest.yml` and `tools/docker/docker-compose.yaml` do similar setup for postgres & minio,
the github version doesn't need quay login (and uses github caches), the compose version does (`docker login quay.io`).

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -76,4 +76,4 @@ jobs:
           METRICS_UTILITY_BUCKET_REGION: "us-east-1"
           METRICS_UTILITY_BUCKET_SECRET_KEY: "myusersecretkey"
         run: |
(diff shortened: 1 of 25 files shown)


Ladas (approved):
👍

MilanPospisil (approved):
LGTM :+1:

PR #75, opened 2025-02-26
AAP-40683 - Indirectly managed nodes

MilanPospisil:
[AAP-40683]

### Supporting documentation: https://github.com/ansible/handbook/blob/main/The%20Ansible%20Engineering%20Handbook/docs/AAP/Services/Metrics%20Utility/CCSP/CCSP.md

Maybe its good idea to read docs first for better understanding how original build_report works (focus only on job hosts summary, events are not needed in this PR).

### Added testing data:
Sql dump for indirectly managed nodes, data can be generated alternatively following this guidelines:
https://docs.google.com/document/d/1_dw_YTwtfRSWuroMfddg5v16m2wx8sB6c52Cl5a0blM/edit?tab=t.0

Dont forget to clean your docker properly before rebuild, because migrations need to run correctly and also, feature is
now merged, so you can use latest devel branch

Also, after playbook will run, indirect table will be empty until data collecting task will run eventually, it can be forced in connecting to python process and running the tasks directly, which is described in the document somewhere in later pages.

Make sure you will not collect data and rewrite data in repository, because data from this PR are needed for testing reports (problem below). So for collecting, use another path in variable METRICS_UTILITY_SHIP_PATH. 

### Potential problem in Controller with host name localhost

host name for indirect hosts stored in controller DB is localhost, but it should be derived from cannonical facts instead and should not be localhost. Ladas told me that this is probably error in controller side. Until repaired, I manualy edited file inside tarball and renamed host name from localhost to something unique, because with localhost, the indirect remote virtual machine will be treated like localhost, which is not and aggregated together with direct localhosts, thus we will not test any indirect hosts at all...

### Collector changes

Collectors file adds new collector method that will collect indirectly managed hosts, renames some columns to match the original jobhosts summary table (direct hosts), so we can run one algorithm on both data. Basicaly we are loading from different table, but output should looks as much similar as jobhosts as possible.

The new collector will also generate tarball like previous two with indirects hosts csv file.

### Extractors changes

extractor_directory (and s3) are changed so now they also can recognise indirect hots csv file there. So iter batches now works the same as worked before and is now serving also indirect hosts.

Each iteration of extractor will serve one tarball which contains either job_hosts_summary, or indirect hosts data, it can serve also events data, but those are ignored for our host counting (this is source of inefectiveness, since the insights collector library produces tarballs that are obscure, unless unpacked, we cant guess what is inside only by tarball name...). But it may be configured in metrics utility? Thats question for future.

Also, code has been refactored to not have so much copy pasted code anymore, this was because of sonarclouds complaints (and my complaints too :-) )

### Job Hosts Summary

The host counting algorithm is adjusted so it can also compute indirect hosts. First, it looks if new data are tarball batch of either direct, or indirect hosts (future will also contains edge hosts). It adds new integer column device_type with constants DIRECT=0 and INDIRECT=1, so we can differentiate between rows in dataframe.

Aggregation of potential duplicate with the same host_name will aggregate device_type as MIN function, because if the group containst DIRECT and INDIRECT hosts, group will become DIRECT (0, min value). For edge nodes, they will have even less priority than INDIRECT, so min will work for EDGE=2 as well.

Also, for direct hosts, we have different granuarity of tasks runs columns - OK, Failed and etc, that are merged into task_runs (summed), while indirect hosts have task_runs column naturaly (renamed from count column in controller DB, csv contains task_runs instead of count).

### Managed hosts and Indirect Managed hosts report sheets

Managed hosts sheet now counts only DIRECT nodes, no matter if the indirect counting is ON or OFF, because device_type column will be set to DIRECT for all direct hosts even if indirect couting is off for code simplicity and clarity.

Indirect Managed hosts are doing the same as Managed hosts, but they filter for INDIRECT hosts, but logic for creating sheet is the same.

Also, I rewrited copy pasted code to use function reference:

 if "managed_nodes_by_organizations" in self.optional_report_sheets():
                func = self._build_data_section_usage_by_node_with_org_details
            else:
                func = self._build_data_section_usage_by_node
                
  so it dont have to be there twice.
  
  ### Usage by organizations new columns for indirect hosts
  
  if INCLUDE_INDIRECT:
            agg_dict["indirect_host_runs_unique"] = (
                "host_name",
                lambda x: x[dataframe.loc[x.index, "device_type"] == INDIRECT].nunique()
            )
            agg_dict["indirect_host_runs"] = (
                "host_name",
                lambda x: x[dataframe.loc[x.index, "device_type"] == INDIRECT].count()
            )
            
   This is adding new columns for indirect hosts and is filtering the aggregation for only INDIRECT hosts, code for directs host is also adjusted for this filter with DIRECT hosts only.

## Testing

METRICS_UTILITY_OPTIONAL_COLLECTORS=indirect_nodes
This will also turn off events counting, since event counting is missing there (its default if env var is empty)

python ./manage.py build_report --since=2025-02-25 --until=2025-02-25 --force --verbose

verbose is optional, but its nice to see how data are processed.

### Expected Results

The spreadsheet is generated with new sheet for indirect hosts:

image

And additional columns in Usage by organization:

image

## Required Actions

- [x] Requires documentation updates

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers


sql/indirect.sql
Ladas, line 1: hm, maybe we should also tarball these? Otherwise these will be huge files

metrics_utility/automation_controller_billing/dataframe_engine/dataframe_jobhost_summary_usage.py
```diff
@@ -2,12 +2,20 @@
...
     Base
 
 logger = logging.getLogger(__name__)
 
+DIRECT = 0
+INDIRECT = 1
+# EDGE = 2
+
+def get_optional_collectors():
+    return os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent').split(",")
```
Ladas: try to pass all the env vars via https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/management/commands/build_report.py#L179

So it's easier to manage the config.

metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
```diff
@@ -8,9 +8,18 @@
...
 
+DIRECT = 0
+INDIRECT = 1
+# EDGE = 2
+
+def get_optional_collectors():
+    return os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent').split(",")
+
+environs = get_optional_collectors()
+INCLUDE_INDIRECT = ('indirect_nodes' in environs)
```
Ladas: similar try to pass all env vars via https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/management/commands/build_report.py#L179

so we don't handle this in each individual file

MilanPospisil: Yeah, I know, I had in in helpers, but import was not working, so I wanted to solve this in morning, so I have pushed my code.

metrics_utility/automation_controller_billing/dataframe_engine/dataframe_jobhost_summary_usage.py
MilanPospisil: I included it here, but I also have helper file with coputing directly variable INCLUDE_INDIRECT so I can easily to import it anywhere, because it is also used in collectors and not only in build_report.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -302,3 +299,52 @@ def main_jobevent_table(since, full_path, until, **kwargs):
...
+            LEFT JOIN main_unifiedjob
+                ON main_unifiedjob.id = main_indirectmanagednodeaudit.job_id
+            LEFT JOIN main_inventory
+                ON main_inventory.id = main_indirectmanagednodeaudit.inventory_id
+            LEFT JOIN main_organization
+                ON main_organization.id = main_unifiedjob.organization_id
+            LEFT JOIN main_unifiedjobtemplate AS main_unifiedjobtemplate_project
+                ON main_unifiedjobtemplate_project.id = main_job.project_id
+            WHERE (main_indirectmanagednodeaudit.created >= '{since.isoformat()}'
+            AND  main_indirectmanagednodeaudit.created < '{until.isoformat()}')
```
himdel, line 342: don't we have some decent way of escaping things?

This is safe as long as they're datetime objects, but otherwise that's how you get sql injections.

(EDIT: but I see the same thing in all the other functions, so, not a thing to fix in this PR :))

Ladas: @himdel  yes, but it's like that in all collectors I think. (including the one in controller) This has only internal access, where you can already access DB creds and have full access then.

That being said, it safer to not have the code sql injectable (although the isoformat will break sql injection attempt). We should just use psycopg, like we do it e.g. in analytics https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/blob/main/restapi/v1/content_explorer/handlers_content_explorer.py#L47

See https://www.psycopg.org/psycopg3/docs/api/sql.html#module-usage, that offers usage for Literal and indentifier, for correct escaping of queries.

Ladas (approved):
👍 looks great, lets build on this further

PR #59, opened 2025-02-03
Sonarcloud integration

ShaiahWren:
[AAP-39024]

## Description

Properly create and configure the `sonar-project.properties` file.

## Further explaination:

Sonarcloud is running a automatic analysis already, but we also require the dynamic analysis that makes calls to sonarcloud from inside a GH action in order to initiate scans. It's explained here in the [handbook](https://handbook.eng.ansible.com/docs/CICD/Services/sonar-config) and mirrors the eda [config](https://github.com/ansible/eda-server/blob/main/sonar-project.properties). As noted below though, the sonar-project.properties config is only the first step in getting this running; we also need to set up a gh actions workflow and the secret sonar token. But we are waiting on [AAP-39227](https://issues.redhat.com/browse/AAP-39227) and in turn [AAP-39220](https://issues.redhat.com/browse/AAP-39220) for that.

## Testing

See /metrics-utility/sonar-project.properties

### Prerequisites

None

## Required Actions

- [ ] Requires documentation updates

- [ ] Requires downstream repository changes

- [ ] Requires infrastructure/deployment changes

- [ ] Requires coordination with other teams

- [ ] Blocked by PR/MR: #XXX

### Code Quality

- [x] Code is properly linted and formatted.
- [ ] All tests (existing and new) pass successfully.
- [ ] Tests are added or updated as needed.
- [ ] Code includes relevant comments for complex sections.
- [ ] Changes are reviewed and approved by at least two team members.
- [ ] Documentation is updated where applicable.  
       - If updates are required in the [handbook](https://github.com/ansible/handbook), create a separate PR for the handbook repo.

## Notes for Reviewers

This pr does _not_ include the following steps:
- GitHub Actions workflow includes steps for SonarCloud scanning and coverage reporting.
- The secret SONAR_TOKEN is configured in the repository.

Anything related to running tests will rely on [AAP-39227](https://issues.redhat.com/browse/AAP-39227) which in turn relies on [AAP-39220](https://issues.redhat.com/browse/AAP-39220).

PR #76, opened 2025-02-26
Remove duplicate sonarcloud workflow

himdel:
The original `.github/workflows/sonarcloud.yml` was part of #59, but had a syntax error, so didn't do anything:

> Invalid workflow file: .github/workflows/sonarcloud.yml#L19
> You have an error in your yaml syntax on line 19

(https://github.com/ShaiahWren/metrics-utility/actions/runs/13162555776)

This was fixed in #72, but that causes the workflow to run - and fail:

> ERROR You are running CI analysis while Automatic Analysis is enabled. Please consider disabling one or the other.

(https://github.com/ansible/metrics-utility/actions/runs/13551856688/job/37877096554?pr=74)

=> Since I'm not seeing any successful "SonarCloud scan" or "SonarCloud Scan" runs anywhere,
and sonarqube works anyway, before and after the change,
I conclude the file is indeed duplicate and not needed, removing.


MilanPospisil:
Yeah, I noticed that warning too.

MilanPospisil (approved):
LGTM :+1:

PR #80, opened 2025-03-03
Tweak sonar duplication checks

Ladas:
/test dir should be ignored, since tests are supposer to be duplicated since tests are not supposed to follow DRY principle

/report dir will be ignored now, since DSL for creating sheets has duplicate code


himdel (approved):
:crossed_fingers:

PR #84, opened 2025-03-06
compose.yml updates

himdel:
* compose complains about deprecated version - remove
* quotes around environment values
* volumes - tweak to fix minio when run the second time 
* sync minio-data name between container and ci
* update s3 commands - `mc` complains about `user svcacct add` vs `accesskey create`; drop usr/bin
* make `metrics-utility` & `metrics-utility-env` containers optional
  * `metrics-utility` will run pytest when invoked with `--profile=pytest`
  * `metrics-utility-env` will run an env to run pytest in manually, when invoked with `--profile=env`
  * but neither gets run by default
* user=1001 - fixes uv run pytest permission denied

Related to AAP-38819 & AAP-39221


Ladas (approved):
Excellent work!

PR #85, opened 2025-03-06
Test gather with directory output

himdel:
Issue: AAP-39221
Follows #52 , #84 

Adds a minimal gather test - this does no checking of the tarball contents, just tests that a set of tarballs was produced and the command ran successfully.

(Follow up PRs for s3 gather, crc push, and inspecting contents still wip.)


.github/workflows/pytest.yml
```diff
@@ -68,6 +68,13 @@ jobs:
           mc admin policy attach local readwrite --user=mynewuser
           mc admin accesskey create local mynewuser --access-key myuseraccesskey --secret-key myusersecretkey
 
+      - name: "Patch mock_awx postgres config"
+        run: |
+          sed -i '/USER/s/myuser/awx/' mock_awx/settings/__init__.py
+          sed -i '/PASSWORD/s/mypassword/awx/' mock_awx/settings/__init__.py
+          echo \\l | sudo su - postgres -c psql
+          echo \\dt | sudo su - postgres -c psql
+
```
AlexSCorey, line 76: I guess its unclear to me why we need this.

tools/docker/roles.sql
```diff
@@ -1 +1 @@
-CREATE ROLE awx;
+CREATE ROLE awx SUPERVISOR LOGIN PASSWORD 'awx';
```
AlexSCorey: Should SUPERVISOR actually be SUPERUSER?
```suggestion
CREATE ROLE awx SUPERUSER LOGIN PASSWORD 'awx';
```

metrics_utility/test/gather/test_directory_gather.py
```diff
@@ -0,0 +1,47 @@
...
+
+file_glob = f'./metrics_utility/test/test_data/data/{year}/*/*/{uuid}-*.tar.gz'
+
+
+def validate_exists(file_glob):
+    assert len(glob.glob(file_glob)) > 0
+
+
+@pytest.fixture
+def cleanup_glob():
```
AlexSCorey, line 26: When does this get called?

tools/docker/roles.sql
AlexSCorey: Change SUPERVISOR to SUPERUSER?
```suggestion
CREATE ROLE awx SUPERUSER LOGIN PASSWORD 'awx';
```

.github/workflows/pytest.yml
himdel: It should eventually be configurable, but right now, the DB test gets the DB credetnials from that file.

But those are different when using the docker compose env vs github CI, so this patches the config to match reality.

metrics_utility/test/gather/test_directory_gather.py
himdel: It gets called around the `test_command` test (because of the `cleanup_glob` param),
so the bits after the `yield` get called after the test has run, anything before the yield would get called before.

(because `@pytest.mark.parametrize('cleanup',` seems to skip the test when the list of files to clean up is empty, which is problematic with globs)

AlexSCorey (approved):
Looks good to me.

PR #89, opened 2025-03-14
AAP-39225 : Github workflow to gate for code coverage in tests.

AlexSCorey:
AAP-39225

## Description

This PR adds a github action which will kick off when a pr is created, or when a pr is updated.  This GHA will check that at least 80% of the code is tested in our CI tests.  

This is important for continued code health, and to help us ensure we are always green.

### Steps to Test

1. Create a PR against this branch.  
2. The GHA should automatically start, and send the results to sonar cloud

### Expected Results

After the coverage check is run a report will be sent to sonar cloud, and the result will be displayed in the actions tab in the repo.  If the 80% threshold is not reached the check should fail blocking the pr from merging.

### Code Quality

## Notes for Reviewers

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -20,9 +20,12 @@ jobs:
 
       - name: "Check for changes in uv.lock"
         run: |
-          # fail if uv sync had to change uv.lock
           uv sync
-          git diff --exit-code uv.lock
+          if ! git diff --quiet uv.lock; then
+            echo "uv.lock has changed. Run 'uv sync' locally and commit the updated lock file."
+            git diff uv.lock
+            exit 1  # Fail the job
+          fi
 
       - name: "Run ruff check"
         run: |
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -74,7 +74,11 @@ jobs:
           sed -i '/USER/s/myuser/awx/' mock_awx/settings/__init__.py
           sed -i '/PASSWORD/s/mypassword/awx/' mock_awx/settings/__init__.py
 
-      - name: "Run pytest"
+      - name: Run pytest
+        run: uv run pytest -s -v  --cov=. --cov-report=xml
+
+      - name: Send code coverage report to Sonar Cloud
+        uses: SonarSource/sonarqube-scan-action@v4
         env:
           METRICS_UTILITY_BUCKET_ACCESS_KEY: "myuseraccesskey"
           METRICS_UTILITY_BUCKET_ENDPOINT: "http://localhost:9000"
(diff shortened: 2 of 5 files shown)


metrics_utility/automation_controller_billing/report/base.py
```diff
@@ -41,7 +41,15 @@ def convert_cell(self, cell):
...
         return cell
+    def _fix_event_host_sample(self, mapping_dataframe, destination_dataframe):
+        if destination_dataframe is None:
+            return None
 
+        def concatenate_columns_mapping(row):
+            return f"{row['original_host_name']}__{str(row['install_uuid'])}__{str(row['job_remote_id'])}"
+
+        def concatenate_columns_destination(row):
+            return f"{row['host_name']}__{str(row['install_uuid'])}__{str(row['job_remote_id'])}"
```
AlexSCorey: This will all be deleted pre-merge.  I added it just to get the coverage report to run.

.github/workflows/pytest.yml
```diff
@@ -74,13 +74,47 @@ jobs:
...
+        id: extract_coverage
+        run: |
+          COVERAGE=$(grep -oP '(?<=line-rate=")[0-9.]+(?=")' coverage.xml)
+          COVERAGE_PERCENT=$(awk "BEGIN {printf \"%.2f\", $COVERAGE * 100}")
+          echo "COVERAGE_PERCENT=$COVERAGE_PERCENT" >> $GITHUB_ENV
+          echo "Code coverage: $COVERAGE_PERCENT%"
+          echo "Coverage: $COVERAGE"
+
+          if (( $(echo "$COVERAGE_PERCENT < 58" | bc -l) )); then
+          echo "::error file=coverage.xml,line=1::Code coverage is below 59% ($COVERAGE_PERCENT%)"
```
AlexSCorey: This throws an error and exits with a code 1 if the overall code coverage falls below 59%, which is where it is currently.

The coverage report coming from Sonar Cloud only takes into account new code, where as this metric takes all code in the valid directories into account.

I left this in as a  measure of reassurance so that our code coverage does not get worse.

.github/workflows/pr-checks.yml
```diff
@@ -20,9 +20,12 @@ jobs:
...
         run: |
-          # fail if uv sync had to change uv.lock
-          uv sync
-          git diff --exit-code uv.lock
+           uv sync
+            if ! git diff --quiet uv.lock; then
+              echo "uv.lock has changed. Run 'uv sync' locally and commit the updated lock file."
+              git diff uv.lock
+              exit 1  # Fail the job
+            fi
```
himdel:
```suggestion
          uv sync
          if ! git diff --quiet uv.lock; then
            echo "uv.lock has changed. Run 'uv sync' locally and commit the updated lock file."
            git diff uv.lock
            exit 1  # Fail the job
          fi
```

(just a surprising indent level, works either way :))

himdel:
Still looks good :+1: 

Remember the `pull_request_target` too before merging.

.github/workflows/pytest.yml
```diff
@@ -74,13 +74,43 @@ jobs:
...
+      - name: Run pytest
+        run: uv run pytest -s -v  --cov=. --cov-report=xml --cov-branch
+
+
+      - name: Extract coverage percentage
+        id: extract_coverage
+        run: |
+          COVERAGE_RATE=$(grep -oP '(?<=line-rate=")[0-9.]+(?=")' coverage.xml)
+
+          if (( $(echo "$COVERAGE_RATE * 100 < 59" | bc -l) )); then
```
Ladas: so all this is not needed when using sonar cloud?

So we should just change the noar-project.properties? 

(sonar seems to be reporting the code coverage must be bigger than 80%)

AlexSCorey: Yes, sonar cloud defaults to 80% code coverage threshold. I haven't yet found a way to customize this value.  There doesn't seem to be a parameter available for us to change that.  It seems that sonar cloud only reports to coverage percentage based on new code only, not existing code.  So, the code mentioned above is a safety net to ensure that our overall coverage does not drop below 59%, which is where it is now.  

That said, I'm sort of indifferent as to whether we keep this in.

himdel (approved):

Ladas (approved):

PR #91, opened 2025-03-18
AAP-41409 - Data generator script

himdel:
Issue: AAP-41409 (related to AAP-41410)

Adds `generator.py`: generate new csv by multiplying rows and applying rules
    
* loads all tarballs from a `SOURCE_DATA_PATH` glob
* loads all csvs in `SELECTED_DATA` (comma-separated list) (and last `config.json`)
* concatenates the data from all tarballs together (each kind of csv as one dataframe)
* for each table
   * repeats each row until there are at least `MAIN_{JOBHOSTSUMMARY,HOST,INDIRECT,JOBEVENT}_UNIQUE_SIZE` items
   * randomizes hostname fields
   * repeats each row until there are at least `MAIN_{JOBHOSTSUMMARY,HOST,INDIRECT,JOBEVENT}_SIZE` items
   * crops row count down to `MAIN_{JOBHOSTSUMMARY,HOST,INDIRECT,JOBEVENT}_SIZE` items
   * recomputes sequential ids
   * randomizes date fields, staying within an interval between `OUTPUT_DATE_FROM` and `OUTPUT_DATE_TO`
* converts back to csv
* saves to `OUTPUT_DATA_PATH` (`$OUTPUT_DATA_PATH/Y/M/D/*.tar.gz`)

(Supports `--help` to see the env var names, `--verbose` to see what's happening.)

(Doesn't know how to split tars, but, do we need to?)

---

```
$ uv run ./generator.py --help
usage: generator [-h] [-v]

options:
  -h, --help     show this help message and exit
  -v, --verbose

Environment vars:
    MAIN_JOBHOSTSUMMARY_SIZE (default: 10000)
    MAIN_JOBHOSTSUMMARY_UNIQUE_SIZE (default: 2000)
    MAIN_HOST_SIZE (default: 10000)
    MAIN_HOST_UNIQUE_SIZE (default: 2000)
    MAIN_INDIRECT_SIZE (default: 10000)
    MAIN_INDIRECT_UNIQUE_SIZE (default: 2000)
    MAIN_JOBEVENT_SIZE (default: 10000)
    MAIN_JOBEVENT_UNIQUE_SIZE (default: 2000)
    SOURCE_DATA_PATH (default: ./metrics_utility/test/test_data/data/{year}/**/*.tar.gz)
    OUTPUT_DATA_PATH (default: ./metrics_utility/test/test_data/data/)
    INPUT_DATE_FROM (default: lastyear-01-01)
    INPUT_DATE_TO (default: year-01-01)
    OUTPUT_DATE_FROM (default: year-01-01)
    OUTPUT_DATE_TO (default: nextyear-01-01)
    SELECTED_DATA (default: job_host_summary,main_host,main_indirectmanagednodeaudit,main_jobevent)
```

Example output: https://docs.google.com/spreadsheets/d/1ohWtDmgKOqRFDc1YWk545FRvITSbLg4S/edit?usp=sharing&ouid=112307144989328414652&rtpof=true&sd=true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+help:
+	@echo help sync test coverage lint fix
+
+sync:
+	uv run sync
+
+test:
+	uv run pytest -s -v
+
+coverage:
+	uv run pytest -s -v --cov=. --cov-report=html
+
+lint:
+	uv run ruff check
+	uv run ruff format --check
+
+fix:
+	uv run ruff check --fix
+	uv run ruff format
+
+.PHONY: help sync test coverage lint fix
--- a/generator.py
+++ b/generator.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+import argparse
 import datetime
 import glob
 import io
@@ -139,9 +140,13 @@ def process_tarballs(path, temp_dir, enabled_set):
 
 class Main:
     def __init__(self):
-        self.set_config()
+        self.parse_env()
+        self.parse_args()
 
-    def set_config(self):
+        if self.verbose:
+            print('config', vars(self))
+
+    def parse_env(self):
         year = datetime.datetime.now(tz=datetime.timezone.utc).year
 
         # data_collection_status = ()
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,21 @@
...
+	@echo help sync test coverage lint fix
+
+sync:
+	uv run sync
+
+test:
+	uv run pytest -s -v
+
+coverage:
+	uv run pytest -s -v --cov=. --cov-report=html
```
AlexSCorey, line 11: our sonar-project properties expects the coverage report to be a `.xml` file.  
```suggestion
	uv run pytest -s -v --cov=. --cov-report=xml
```

generator.py
```diff
@@ -0,0 +1,219 @@
...
+def random_date(earliest, latest):
+    tsmin = earliest.timestamp()
+    tsmax = latest.timestamp()
+    rand = tsmin + (random.random() * (tsmax - tsmin))
+    return datetime.datetime.fromtimestamp(rand, tz=datetime.timezone.utc)
+
+
+def random_hostname():
+    nouns = 'armadillo axolotl badger beetle bison buffalo capybara caribou cassowary chameleon cheetah cobra coyote dolphin eagle elephant falcon ferret flamingo fox gazelle giraffe hippo ibex jaguar kangaroo koala lemur leopard lynx macaw meerkat narwhal octopus orangutan otter owl panda panther peacock pelican penguin pigeon puma rabbit raven rhino sparrow tiger toucan turtle whale wolf wombat zebra'.split(
+        ' '
```
AlexSCorey: I wouldn't have thought of this approach.  I like it!

AlexSCorey (commented):
I know this is still a draft.  I couldn't help myself from looking at it.

Makefile
himdel: would we want to do that outside the context of uploading to sonarcloud?

I was thinking of this as a Makefile for commands a dev would be running manually, I can see how to use a html report but what would we do with an xml one?

generator.py
himdel: thanks :)

I started from a password generator (https://github.com/himdel/dotfiles/blob/master/bin/genpass), but the hostnames looked too weird :)

It only creates about 2000 unique word combinations, but, times 900 numeric should be enough for our purpose :)

Makefile
AlexSCorey: So my primary concern here is that over time we will begin to use the make commands to execute work in our CI or other automations.  If someone uses this command to run the tests in automation expecting that they will get a coverage report from sonar cloud they will be mistaken.  XML looks a lot like html, and can be viewed in a browser just like html.  It also reads in the code ver similarly to html.  

This isn't a merge blocker for me but is something we need to be cognizant of if this pr gets merged as is.

AlexSCorey:
I noticed that there is no code coverage reported.  Is this branch up-to-date with devel?

AlexSCorey (approved):
All of my concerns have been addressed.  Well done!

Ladas (approved):
looks great 👍

Makefile
himdel: Yeah, agreed, let's not do that, and stop anyone from doing it doing reviews :)

The html output is very different from a xml one, I don't think you can easily confuse the two, but if anyone wants to read xml, they'll have to run it manually :)

PR #93, opened 2025-03-19
Fixup combine_* function naming

himdel:
Fixup function naming,
sync names `set_merge`,`dict_merge`,`dict_set_merge` -> `combine_set`,`combine_json`,`combine_json_values`
related to #78.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,7 +1,7 @@
 name: "Pytest tests"
 
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   pytest:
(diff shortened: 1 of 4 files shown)


AlexSCorey (approved):
This looks good to me.  I do think we want to rerun checks so we can get ensure the code coverage doesn't drop.

himdel:
Oh, well, it fails after rebase, because of that `pull_request_target` :) 
(https://github.com/ansible/metrics-utility/actions/runs/13977615413/job/39135150900?pr=93#step:13:50)

Adding here :) (=> related to #89)
We'll see in the next PR :)

PR #94, opened 2025-03-19
Start 0.4.2 changelog, fixup pglock import

himdel:
(Was #83, pre-rebase)

These are changes originally planned for a 0.4.2 release, which we didn't end up doing.
Porting these to current devel:

* starts a 0.4.2 (unreleased) changelog entry
* updates the versions everywhere
* fixes the `advisory_lock` import - checking for django-ansible-base is problematic in early 2.5 which has dab, but not a version with the change - just try/catch it

This should fix tests using 2.5 controller with nightly metrics util.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,3 +54,6 @@
 
 ## 0.4.1
 - relax boto3 requirement so it builds with any version available
+
+## 0.4.2 (unreleased)
+- allow importing pglock from either awx or django-ansible-base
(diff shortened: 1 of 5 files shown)


AlexSCorey (approved):
This looks good to me.  Seems like we might want to track removing this code and just importing pglock from DAB in future major releases, right?

himdel:
Thanks :)

> Seems like we might want to track removing this code and just importing pglock from DAB in future major releases, right?

In theory, yes, though, given the extended support for 2.4, we probably can't actually do it until 2027(?), or until we have to split into 2.4, 2.5 and 2.6 branches. Or, until we're switched to a version-dependent service. I hope we can get rid of importing any awx code before then :)

PR #95, opened 2025-03-19
Tests - bump coverage by running simple tests in both modes

himdel:
Follows #89, AAP-39225

Move the logic of running the command from tests - internally or externally - to a util file,
always call `run_build_ext`, `run_build_int`, `run_gather_ext` or `run_gather_int`.

The `_ext` variants use a subprocess, and take `(env dict, args array)`,
the `_int` variants use a `Command` import, and take `(env dict, options dict)`.

And updates every test file that only called the `_ext` variant to call both.
This *should* increase coverage because the `_ext` variant doesn't collect it (=> related to #89).

Importing the gather command fails from tests, because awx is not mocked - call metrics utility `prepare()` to set up the mocks even in tests.

Also updates tests running metrics-utility as a command to explicitly output the command stderr, instead of seeing it as part of an exception - fixes escaped newlines :)

(And gather doesn't do `--force`, removed from tests.)

### DIFF
--- a/metrics_utility/test/ccspv_reports/test_CCSP.py
+++ b/metrics_utility/test/ccspv_reports/test_CCSP.py
@@ -1,11 +1,10 @@
-import subprocess
-import sys
-
 from datetime import datetime
 
 import pytest
 
 from conftest import validate_sheet_columns, validate_sheet_tab_names
+
+from metrics_utility.test.util import run_build_ext, run_build_int
 
 
 env_vars = {
@@ -20,7 +19,6 @@ env_vars = {
     'METRICS_UTILITY_REPORT_SKU': 'MCT3752MO',
     'METRICS_UTILITY_REPORT_EMAIL': 'email@email.com',
     'METRICS_UTILITY_REPORT_TYPE': 'CCSP',
-    'AWX_LOGGING_MODE': 'stdout',
 }
 
 file_path = './metrics_utility/test/test_data/reports/2024/02/CCSP-2024-02.xlsx'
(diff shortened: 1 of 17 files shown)


himdel:
Locally generated html coverage report seems to jump from 59% to 70% :tada:

metrics_utility/test/ccspv_reports/test_CCSP.py
```diff
@@ -224,16 +223,29 @@
...
 
-    python_executable = sys.executable
-    result = subprocess.run(
-        [python_executable, 'manage.py', 'build_report', '--month=2024-02', '--force'],
-        text=True,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        env=env_vars,
-    )
+    run_build_ext(env_vars, ['--month=2024-02', '--force'])
```
AlexSCorey, line 227: what is `ext` meant to be in this case?

metrics_utility/test/gather/test_s3_gather.py
```diff
@@ -12,23 +13,25 @@
     'METRICS_UTILITY_REPORT_TYPE': 'CCSPv2',
     'METRICS_UTILITY_SHIP_PATH': 'metrics-utility/shipped_data',
     'METRICS_UTILITY_SHIP_TARGET': 's3',
-    'AWX_LOGGING_MODE': 'stdout',
 }
 
 
+# mc ls -r local/metricsutilitys3/metrics-utility/shipped_data/data/
```
AlexSCorey: Can this be removed?

AlexSCorey (commented):
Really just some small questions seeking clarity, though we do need to track down why your sonar cloud code coverage is reporting 0%.

metrics_utility/test/ccspv_reports/test_CCSP.py
himdel: external .. (vs int for internal) .. but open to better names :) (cli vs command didn't sound clear enough :))

himdel:
> why your sonar cloud code coverage is reporting 0%

I assume it's because I'm not adding any code other than tests.

But if you run coverage manually, you'll see 59% before and 70% after.

himdel:
Rebased,
html coverage report before: [56%](https://himdel.eu/htmlcov-devel-20250327/)
html coverage report after: [67%](https://himdel.eu/htmlcov-pr95-20250327/)

MilanPospisil (approved):
LGTM :+1 Good job, CI test passing, code makes sense :-)

PR #96, opened 2025-03-24
Allow python 3.13

himdel:
We had a python < 3.13 requirements because pandas was failing to compile, `C++17: error: standard attributes in middle of decl-specifiers (CYTHON_UNUSED)`.

Fixed in newer pandas, bumping.


AlexSCorey (approved):
Looks good.

PR #90, opened 2025-03-18
Format

himdel:
This just runs `uv run ruff format` on the codebase,
and enables the format pr-check.

PR #53, opened 2025-01-29
AAP-39264: add helpful settings to ruff linter

ShaiahWren:
[AAP-39264]

## Description

Add additional settings to the existing ruff linter in order to enhance code quality and readability. In tandem with https://github.com/ansible/metrics-utility/pull/49.

## Code changes

- Prefer single quotes over double quotes. (The initial black formatter had `skip-string-normalization = true` to not do any normalization )
```
[tool.ruff.format]
quote-style = "single"
```
- Import sorting - separate direct and 'from' imports
```
[tool.ruff.lint.isort]
lines-between-types = 1
```
- Import sorting - order by case and alphabetically
```
[tool.ruff.lint.isort]
order-by-type = true
```
- Import format - Use a single line after each import block.
```
[tool.ruff.lint.isort]
lines-after-imports = 1
```

### Steps to Test

1. Run `ruff check` on /metrics-utility

### Expected Results
All checks should pass

## Required Actions

- [ ] Requires documentation updates

- [ ] Requires downstream repository changes

- [ ] Requires infrastructure/deployment changes

- [ ] Requires coordination with other teams

- [ ] Blocked by PR/MR: #XXX

### Code Quality

- [x] Code is properly linted and formatted.
- [x] All tests (existing and new) pass successfully.
- [ ] Tests are added or updated as needed.
- [x] Code includes relevant comments for complex sections.
- [x] Changes are reviewed and approved by at least two team members.
- [ ] Documentation is updated where applicable.  
       - If updates are required in the [handbook](https://github.com/ansible/handbook), create a separate PR for the handbook repo.

## Notes for Reviewers
None

PR #97, opened 2025-03-24
Enable isort

himdel:
Related to #53, AAP-39264
Follows #90 

We had config for `ruff.lint.isort` to sort imports, but it was not enabled in `ruff.lint extend-select` - enabling.

But updated `lines-after-imports` to 2, which causes `ruff check` and `ruff format` to agree, whereas with 1, check would remove a line only for format to add it. (And `order-by-type` is already true by default.)


pyproject.toml
```diff
@@ -55,12 +55,12 @@ extend-select = [
 #   "N",
 # Python syntax upgrades
 #   "UP",
+# Sort imports
+    "I",
```
AlexSCorey, line 59: What does this line do?

AlexSCorey (approved):
Looks good to me.

pyproject.toml
himdel: This is the thing that makes ruff not ignore isort-related rules.

I haven't found a good legend for what each letter means, but, `I` is `isort` :)

himdel: Found it :)

```
$ uv run ruff linter 
   F Pyflakes
 E/W pycodestyle
 C90 mccabe
   I isort
   N pep8-naming
   D pydocstyle
  UP pyupgrade
 YTT flake8-2020
 ANN flake8-annotations
ASYNC flake8-async
   S flake8-bandit
 BLE flake8-blind-except
 FBT flake8-boolean-trap
   B flake8-bugbear
   A flake8-builtins
 COM flake8-commas
 CPY flake8-copyright
  C4 flake8-comprehensions
 DTZ flake8-datetimez
 T10 flake8-debugger
  DJ flake8-django
  EM flake8-errmsg
 EXE flake8-executable
  FA flake8-future-annotations
 ISC flake8-implicit-str-concat
 ICN flake8-import-conventions
 LOG flake8-logging
   G flake8-logging-format
 INP flake8-no-pep420
 PIE flake8-pie
 T20 flake8-print
 PYI flake8-pyi
  PT flake8-pytest-style
   Q flake8-quotes
 RSE flake8-raise
 RET flake8-return
 SLF flake8-self
SLOT flake8-slots
 SIM flake8-simplify
 TID flake8-tidy-imports
  TC flake8-type-checking
 INT flake8-gettext
 ARG flake8-unused-arguments
 PTH flake8-use-pathlib
  TD flake8-todos
 FIX flake8-fixme
 ERA eradicate
  PD pandas-vet
 PGH pygrep-hooks
  PL Pylint
 TRY tryceratops
 FLY flynt
 NPY NumPy-specific rules
FAST FastAPI
 AIR Airflow
PERF Perflint
FURB refurb
 DOC pydoclint
 RUF Ruff-specific rules
```

also, https://docs.astral.sh/ruff/rules/.

Out of which, `["E4", "E7", "E9", "F"]` are enabled by default (https://docs.astral.sh/ruff/settings/#lint_select),
along with `E, W, I` enabled in our config)

PR #99, opened 2025-03-25
Ensures that indirectly_managed_nodes sheet has the correct columns

AlexSCorey:
AAP-41419

## Description
This PR fixes a bug where the columns for the Indirectly managed nodes sheet were not consistently the same columns.  To resolve this I realized that when we add `managed_nodes_by_organizations` to the optional sheet env var it was forcing the use an a function for the `indirectly_managed_nodes` sheet as well.  This PR resolves that bug, and adds a test to ensure it doesn't return.
## Testing
I added a test that creates to reports with different env vars.  Then the test inspects each report and checks that each column consists in each report thus ensuring consistent columns and data.

### Steps to Test

1.Run the new test.

### Code Quality

### DIFF
--- a/metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
+++ b/metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
@@ -144,7 +144,10 @@ class ReportCCSPv2(Base):
             self.wb.create_sheet(title='Indirectly Managed nodes')
             ws = self.wb.worksheets[sheet_index]
             indirects = job_host_summary_dataframe[job_host_summary_dataframe['managed_node_type'] == INDIRECT]
-            func(1, ws, indirects, managed_node_type='indirect')
+            ## This function creates the correct columns for this sheet.  Using the func variable from above
+            ## can result in the wrong columns for this sheet if `managed_nodes_by_organization`
+            ## exists in the METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS env var.
+            self._build_data_section_usage_by_node(1, ws, indirects, managed_node_type='indirect')
             sheet_index += 1
 
         if 'inventory_scope' in self.optional_report_sheets():
--- a/metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
+++ b/metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
@@ -1,10 +1,10 @@
 from datetime import datetime
 
 import openpyxl
-import pandas
 import pytest
 
-from conftest import temporary_env
+from conftest import temporary_env, transform_sheet
+from pandas import Timestamp, read_excel
 
 from metrics_utility.management.commands.build_report import Command
 
@@ -26,9 +26,7 @@ env_vars = {
     'METRICS_UTILITY_REPORT_EMAIL': 'email@email.com',
     'METRICS_UTILITY_REPORT_TYPE': 'CCSPv2',
 }
-
-indirect_nodes_and_managed_by_org_file_path = './metrics_utility/test/test_data/reports/2025/02/CCSPv2-2025-02-13--2025-02-13.xlsx'
-indirectly_managed_nodes_file_path = './metrics_utility/test/test_data/reports/2025/02/CCSPv2-2025-02-13--2025-02-13.xlsx'
+file_path = './metrics_utility/test/test_data/reports/2025/02/CCSPv2-2025-02-25--2025-02-26.xlsx'
 date_today = datetime.now().strftime('%b %d, %Y')
 
 
@@ -36,7 +34,7 @@ date_today = datetime.now().strftime('%b %d, %Y')
 @pytest.mark.parametrize(
     'cleanup',
     [
-        indirect_nodes_and_managed_by_org_file_path,
+        file_path,
     ],
     indirect=True,
 )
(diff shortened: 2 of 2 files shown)


metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
```diff
@@ -143,7 +143,7 @@ def build_spreadsheet(self):
             self.wb.create_sheet(title='Indirectly Managed nodes')
             ws = self.wb.worksheets[sheet_index]
             indirects = job_host_summary_dataframe[job_host_summary_dataframe['managed_node_type'] == INDIRECT]
-            func(1, ws, indirects, managed_node_type='indirect')
+            self._build_data_section_usage_by_node(1, ws, indirects, managed_node_type='indirect')
```
himdel, line 150: All the other if branches use `func`, based on the logic above, so this looks like a bug. Can you add a comment about it being intentional?

(I think I agree this does the right thing.)

metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
```diff
@@ -0,0 +1,82 @@
...
+    with temporary_env({**env_vars, 'METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS': optional_sheets}):
+        options = {
+            'since': '2025-02-13',
+            'until': '2025-02-13',
+            'ephemeral': None,
+            'force': True,
+            'verbose': False,
+        }
+        command = Command()
+        command.handle(**options)
```
himdel: (if #95 goes in first, this should become `run_build_int`)

metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
```diff
@@ -0,0 +1,84 @@
+from datetime import datetime
```
Ladas: for testing this, you could just copy the test in https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/test/ccspv_reports/test_complex_CCSPv2_with_indirect.py

add the extra sheet breaking it and making sure it passes the same asserts (so you can e.g. extract the asserted values into 1 place, to ensure the content is the same and it doesn't drift apart)

Which might be easier way to test it (while still seeing the content of the sheet and what is changing)

metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
```diff
@@ -0,0 +1,84 @@
...
+    'METRICS_UTILITY_REPORT_END_USER_COUNTRY': 'US',
+    'METRICS_UTILITY_REPORT_COMPANY_NAME': 'Partner A',
+    'METRICS_UTILITY_REPORT_SKU': 'MCT3752MO',
+    'METRICS_UTILITY_REPORT_EMAIL': 'email@email.com',
+    'METRICS_UTILITY_REPORT_TYPE': 'CCSPv2',
+}
+
+indirect_nodes_and_managed_by_org_file_path = './metrics_utility/test/test_data/reports/2025/02/CCSPv2-2025-02-13--2025-02-13.xlsx'
+indirectly_managed_nodes_file_path = './metrics_utility/test/test_data/reports/2025/02/CCSPv2-2025-02-13--2025-02-13.xlsx'
+date_today = datetime.now().strftime('%b %d, %Y')
```
himdel: Both `indirect_nodes_and_managed_by_org_file_path` and `indirectly_managed_nodes_file_path` are the same file, maybe this doesn't have to be 2 vars?

(Or, if it does, only one of them is currently being cleaned up)

And `date_today` is unused, remove? :)

metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
himdel: This is still unresolved, should be

```py
def build_workbook(optional_sheets):
    """Build xlsx report using build command and test its contents."""

    run_build_int({
            **env_vars,
            'METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS': optional_sheets,
        }, {
            'since': '2025-02-13',
            'until': '2025-02-13',
            'ephemeral': None,
            'force': True,
            'verbose': False,
    })
```

(+- formatting :))

metrics_utility/test/ccspv_reports/test_consistent_indirectly_managed_nodes_output.py
```diff
@@ -0,0 +1,166 @@
...
+@pytest.mark.parametrize(
+    'cleanup',
+    [
+        file_path,
+    ],
+    indirect=True,
+)
+def test_command(cleanup):
+    build_workbook('indirectly_managed_nodes')
+    build_workbook('indirectly_managed_nodes,managed_nodes_by_organizations')
```
himdel: why are we building two reports, but ignoring the first one?

AlexSCorey: Thats a good catch.  I've modified the code now.  Should be ready for another look @himdel

himdel (approved):
LGTM,
~~will address https://github.com/ansible/metrics-utility/pull/99#discussion_r2022850390 in a second pr~~ addressed :+1: .

PR #108, opened 2025-04-03
run pytest on push - to collect coverage for the devel branch

himdel:
Right now, we're collecting coverage from PR branches,
but the workflow never runs on `devel` - enabling that :)

Closes #107  (I believe)


SherinV (approved):
Thanks!

PR #110, opened 2025-04-03
Add optional report sheet showing collection coverage

himdel:
Issue: AAP-41887

(Started from #100, kudos to @AlexSCorey)

This adds a new optional "Data collection status" sheet to the CCSPv2 report,
enabled by `METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS=data_collection_status,...`.

The data is based on `data_collection_status.csv` output by the collector.

It lists collection instances (when data was actually gathered), highlighting:
* unusual gaps between collections (based on `collection_start_timestamp`)
* gaps in collected intervals (based on `since` vs `until`)

---

Also, a bit of cleanup/refactoring in reports (first commit):

remove unused (and inaccurate) LOG_PREFIX from a bunch of places
remove copypased (and inaccurate) comments
fix a couple of comments
    
move `_build_data_section_usage_by_node_with_org_details` & `_build_data_section_usage_by_job` from report base to report_ccsp_v2 (only used there)
    
replace `_init_dimensions` for main sheet + same for loop in each `_build_*` function,
move into base `add_sheet(title, sheet_index, widths=None)`
    
Fail with "No billing data" only when all dataframes are empty
([0] may be None if no sheets are using it even when data is available)

---

Example output:

* https://docs.google.com/spreadsheets/d/1ohWtDmgKOqRFDc1YWk545FRvITSbLg4S/edit?gid=1350376622#gid=1350376622 (last sheet)
* https://docs.google.com/spreadsheets/d/1kDx82n1IVF0T4ZY2Iuz2bDh5fJT9A6mV/edit?gid=1669669102#gid=1669669102 (only sheet)

### DIFF
--- a/metrics_utility/automation_controller_billing/report/base.py
+++ b/metrics_utility/automation_controller_billing/report/base.py
@@ -17,6 +17,8 @@ class Base:
     RED_COLOR_HEX = 'FF0000'
     LIGHT_BLUE_COLOR_HEX = 'd4eaf3'
     GREEN_COLOR_HEX = '92d050'
+    YELLOW_WARNING_COLOR_HEX = 'FFCC17'
+
     FONT = 'Arial'
     PRICE_FORMAT = '$#,##0.00'
     HOST_NAME = 'Host name'
@@ -44,6 +46,17 @@ class Base:
         # Otherwise, return the cell unchanged.
         return cell
 
+    def add_sheet(self, title, sheet_index, widths=None):
+        self.wb.create_sheet(title=title)
+        ws = self.wb.worksheets[sheet_index]
+        if widths:
+            self.set_widths(ws, widths)
+        return ws
+
+    def set_widths(self, ws, widths):
+        for key, value in widths.items():
+            ws.column_dimensions[get_column_letter(key)].width = value
+
     def _fix_event_host_names(self, mapping_dataframe, destination_dataframe):
         if destination_dataframe is None:
             return None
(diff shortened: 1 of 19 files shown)


himdel:
```
+ xlsx2csv -a ./metrics_utility/test/test_data//reports/2024/04/CCSPv2-2024-04.xlsx
-------- 1 - Data collection status
Collection timestamp,Filter since,Filter until,CSV filename,Status,Elapsed
2024-04-17 13:59:12.537121+00:00,2024-04-15 00:00:00+00:00,2024-04-16 00:00:00+00:00,job_host_summary.csv,ok,0
2024-04-17 13:59:12.576144+00:00,2024-04-16 00:00:00+00:00,2024-04-17 00:00:00+00:00,job_host_summary.csv,ok,0
2024-04-17 13:59:12.592113+00:00,2024-04-17 00:00:00+00:00,2024-04-17 13:59:12.470584+00:00,job_host_summary.csv,ok,0
```

:tada: :)

metrics_utility/automation_controller_billing/report/report_ccsp.py
```diff
@@ -102,70 +94,58 @@ def build_spreadsheet(self):
...
-                ws = self.wb.worksheets[sheet_index]
+                ws = self.add_sheet('Usage by roles', sheet_index, self.config['data_column_widths'])
                 self._build_data_section_usage_by_roles(1, ws, events_dataframe)
                 sheet_index += 1
 
             if 'usage_by_modules' in self.optional_report_sheets():
                 # Sheet with usage by modules
-                self.wb.create_sheet(title='Usage by modules')
-                ws = self.wb.worksheets[sheet_index]
+                ws = self.add_sheet('Usage by modules', sheet_index, self.config['data_column_widths'])
```
AlexSCorey, line 143: We seem to be using this value in many places in this file.  Perhaps it makes sense to assign it to a variable and use that variable in these if blocks.  Might make future maintenance easier.

metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
```diff
@@ -178,8 +182,100 @@ def build_spreadsheet(self):
...
+
+        for r_idx, row in enumerate(rows, current_row):
+            for c_idx, value in enumerate(row, 1):
+                cell = ws.cell(row=r_idx, column=c_idx)
+                cell.value = value
+            row_counter += 1
+
+        return current_row + row_counter
+
+    def _build_data_section_collection_missing(self, current_row, ws, df):
```
AlexSCorey, line 205: It isn't super clear to me what this function is doing in part.  I think the function name might be confusing.  Could we 1) rename the function and/or 2) add more detail to the comments?

metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
```diff
@@ -178,8 +182,100 @@ def build_spreadsheet(self):
...
+                'until': 'Missing from',
+                'next_since': 'Missing until',
+                'gap': 'Gap in seconds',
+            }
+        )
+
+        rows = dataframe_to_rows(dataframe, index=False)
+        return self._build_table(current_row, ws, rows)
+
+    def _build_data_section_collection_status(self, first_row, ws, df):
```
AlexSCorey, line 233: To build the rows for this file, are we looping through each row, of each file in the tarball?

metrics_utility/automation_controller_billing/report/base.py
```diff
@@ -17,6 +17,8 @@ class Base:
     RED_COLOR_HEX = 'FF0000'
     LIGHT_BLUE_COLOR_HEX = 'd4eaf3'
     GREEN_COLOR_HEX = '92d050'
+    YELLOW_WARNING_COLOR_HEX = 'FFCC17'
```
AlexSCorey: I probably did this, but here is a suggestion.
```suggestion
    YELLOW_COLOR_HEX = 'FFCC17'
```
or
```suggestion
WARNING_COLOR_HEX = 'FFCC17'
```

In either case I think we should be consistent in the naming since I expect we will use these colors more as we grow.  Additionally, I think the colors should match the colors and use cases that [Patternfly](https://www.patternfly.org/design-foundations/colors) uses

AlexSCorey (commented):
All that code clean up feels so good.

AlexSCorey (commented):
I ran the test and the report was created an looks great.  The code looks good, I've just left a few comments, and a few questions.  

I did have some concerns about the top part of the report.  It is a bit confusing to me and I'm not sure about its purpose.
Screenshot 2025-04-15 at 4 15 24 PM

metrics_utility/automation_controller_billing/report/report_ccsp_v2.py
himdel: Not each file, just `data_collection_status.csv`

metrics_utility/automation_controller_billing/report/report_ccsp.py
himdel: It's only used once here, and adding a level of indirection for a string would make things more confusing.
This way, we immediatgely see this is about usage by modules, if we move it to a constant, you'd have to look up the constant every time to tell which sheet it is.

himdel:
Thanks @AlexSCorey, added a comment on the "*missing" function, and fixed the color variable :)

AlexSCorey (approved):
This looks good to me.  Still a bit concerned about the code coverage reporting.

PR #114, opened 2025-04-23
Since,until - better param validation

himdel:
Issue: AAP-44487

This ensure that `--since` and `--until` will error out when passed a bare number,
and adds an error message about supporting `<N>d`, `<N>m` and actual dates.


Ladas (approved):
Looks great. 👍

PR #115, opened 2025-04-24
Collection coverage - notice gaps at beginning,end of interval

himdel:
Issue: AAP-44665
Follows #110 , enabled when `METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS=data_collection_status`

This should address gap detection failing at the start & end of collected interval.

Have to pass in the start&until datetimes to the report for that.

### DIFF
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -68,7 +68,7 @@ class Command(BaseCommand):
 
         # parse params
         opt_month = options.get('month') or None
-        opt_month, month = self._handle_month(opt_month)
+        opt_month, month, next_month = self._handle_month(opt_month)
 
         # Since and ephemenral params are specific to subset of reports
         opt_since = None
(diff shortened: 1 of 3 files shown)


metrics_utility/management/commands/build_report.py
```diff
@@ -208,6 +210,11 @@ def _handle_extra_params(self, ship_target=None):
         )
         return base
 
+    def _next_month(self, date):
+        if date.month == 12:
+            return date.replace(month=1, year=date.year + 1)
+        return date.replace(month=date.month + 1)
```
Ladas: date + relativedelta(months=1) should result in the same?

himdel: oh, thanks! timedelta didn't have months, missed relativedelta :)

Ladas (commented):
looks great, lets just add tests

himdel:
Test added .. and fixed `--until` handling which was off by a day

Ladas (approved):
Looks awesome and nice test!!!

PR #116, opened 2025-04-24
[DO NOT MERGE] Stubs column headers to prevent traceback

AlexSCorey:
AAP-44489
This PR prevents a failure and traceback when a user runs a build_report command without any data.  To prevent the failure we stub out the column headers so the things don't fall over when formatting, and generating a spreadsheet.  This allows us to generate a spreadsheet that has all the proper meta-data (title, labels, column headers etc.) without any controller data.


metrics_utility/automation_controller_billing/dataframe_engine/base.py
```diff
@@ -156,6 +160,15 @@ def unique_index_columns():
...
 
+    @staticmethod
+    def default_data_column_headers(file_name):
+        with tempfile.TemporaryDirectory(prefix='automation_controller_billing_data_') as temp_dir:
+            if os.path.exists(os.path.join(temp_dir, f'{file_name}')):
+                row1 = pd.read_csv(os.path.join(temp_dir, f'{file_name}'), nrows=0)
+                return pd.DataFrame(columns=row1)
+            else:
+                return pd.DataFrame([{}])
+
```
himdel: this will never return anything other than the else dataframe - there is no csv in a freshly created temp dir, and we should not assume there would be one anywhere.

However, this is only called from the dataframe classes, and they each have a `data_columns` method...

metrics_utility/automation_controller_billing/dataframe_engine/dataframe_collection_status.py
```diff
@@ -12,7 +12,7 @@ def build_dataframe(self):
         for date in self.dates():
             for data in self.extractor.iter_batches(date=date):
                 batch = data['data_collection_status']
-                if batch.empty:
+                if batch is None or batch.empty:
```
himdel: (wondering if we need these, since the extractor now returns an empty dataframe - goes for all the None checks and `.get` calls)

metrics_utility/automation_controller_billing/dataframe_engine/dataframe_jobhost_summary_usage.py
```diff
@@ -148,7 +150,7 @@ def sum_reachable_columns(row):
                 print_data(billing_data_monthly_rollup, 'Actual global data')
 
         if billing_data_monthly_rollup is None:
-            return None
+            return self.default_data_column_headers('job_host_summary.csv')
```
himdel:
```suggestion
            return pd.DataFrame(columns=self.data_columns())
```

metrics_utility/automation_controller_billing/dataframe_engine/db_dataframe_host_metric.py
```diff
@@ -34,6 +34,6 @@ def build_dataframe(self):
                 host_metric_concat = pd.concat([host_metric_concat, host_metric], ignore_index=True)
 
         if host_metric_concat is None:
-            return None
+            return self.default_data_column_headers('job_host_summary.csv')
```
himdel:
```suggestion
            return pd.DataFrame(columns=self.data_columns())
```

(or at least not `job_host_summary` - should be `host_metric`?)

metrics_utility/automation_controller_billing/extract/base.py
```diff
@@ -101,6 +100,15 @@ def build_data_batch(self, temp_dir, file_name):
...
+            self.default_column_headers(temp_dir, file_name)
+
+    @staticmethod
+    def default_column_headers(temp_dir, file_name):
+        # if file_name is 'main_indirectmanagednodeaudit':
+        if os.path.exists(os.path.join(temp_dir, f'{file_name}.csv')):
+            row1 = pd.read_csv(os.path.join(temp_dir, f'{file_name}.csv'), nrows=0)
+            return pd.DataFrame(columns=row1.columns)
         else:
             return pd.DataFrame([{}])
```
himdel: The extractor is called by the specific dataframe class - so this is not needed, the dataframe can always pass in its own `data_columns` to the extractor

Or, we could just return None in the extractor, in which case the None checks in the dataframes make sense .. and then handle it only in the dataframe.

(EIther way, this needs to work even when the csv is missing/broken...)

himdel (changes requested):
Don't use `read_csv` to figure out the columns, use `data_columns`,
and since we're moving from None to empty (and not the other way around), I assume the new extra None checks are superfluous?

metrics_utility/automation_controller_billing/dataframe_engine/db_dataframe_host_metric.py
```diff
@@ -34,6 +34,6 @@ def build_dataframe(self):
                 host_metric_concat = pd.concat([host_metric_concat, host_metric], ignore_index=True)
 
         if host_metric_concat is None:
```
himdel:
```suggestion
        if host_metric_concat is None or host_metric_concat.empty:
```

(and the others too - otherwise we can still return an empty dataframe without columns)

himdel:
Since the dataframe iter batches loop always skips empty dataframes on input, they should never actually end up returned by build_dataframe...

So, if the only change was in the dataframes classes, from

```py
 if THINGY_rollup is None:
    return None
```

to

```py
if THINGY_rollup is None or THINGY_rollup.empty:
    return pd.DataFrame(columns=self.data_columns())
```

wouldn't that be all that's needed to ensure the value is never None and always has columns?

(Trying in #118)

metrics_utility/test/ccspv_reports/test_CCSPv2_with_no_data.py
```diff
@@ -0,0 +1,50 @@
...
+    'METRICS_UTILITY_REPORT_H1_HEADING': 'CCSP NA Direct Reporting Template',
+    'METRICS_UTILITY_REPORT_PO_NUMBER': '123',
+    'METRICS_UTILITY_SHIP_TARGET': 'directory',
+    'METRICS_UTILITY_REPORT_COMPANY_NAME': 'Partner A',
+    'METRICS_UTILITY_REPORT_SKU': 'MCT3752MO',
+    'METRICS_UTILITY_REPORT_EMAIL': 'email@email.com',
+    'METRICS_UTILITY_REPORT_TYPE': 'CCSPv2',
+}
+
+file_path = './metrics_utility/test/test_data/reports/2024/02/CCSPv2-2024-01.xlsx'
```
himdel:
```suggestion
file_path = './metrics_utility/test/test_data/reports/2024/01/CCSPv2-2024-01.xlsx'
```

mismatch between the month in the filename & the month dir

PR #118, opened 2025-04-28
CCSP* `build_dataframe`: don't return None on empty

himdel:
Related to AAP-44489, #116

Make sure that a `Dataframe*` class `build_dataframe` will never return None,
returning an empty dataframe with known columns instead.

That way, code doing operations like `dataframe['host_name'].nunique()` won't fail on `NoneType` or on unknown columns.

With tests from #119 to test each sheet can be rendered on its own, with no data, empty data, etc


himdel:
(switched from `subprocess.run` -> `run_build_ext` -> `run_build_int`;
with run build ext coverage was 69.0%,
with run build int it's 72.4%)

Ladas (approved):
Looks great!!!

PR #121, opened 2025-04-29
Change version to 0.5.0

himdel:
Issue: AAP-38426

Fixup the about-to-be-released version number ... assumed 0.4.2 but the Issue mentions 0.5.0,
and 0.4.2 sounds too 2.4-like, or 2024-like, not doing that ..  bumping version to 0.5.0 instead.

This corresponds to 2.5.20250507.


Ladas (approved):
👍

PR #120, opened 2025-04-28
Update CHANGELOG for 0.5.0

himdel:
A CHANGELOG update for ~~0.4.2~~ 0.5.0 (fixup in #121)  :tada: 

---

This is a condensed version of what github would generate, which is...

```
## What's Changed
* Add Pull Request Template by @MilanPospisil in https://github.com/ansible/metrics-utility/pull/35
* Fix CI  by @ZitaNemeckova in https://github.com/ansible/metrics-utility/pull/38
* Add testing data to gitignore by @MilanPospisil in https://github.com/ansible/metrics-utility/pull/43
* Test Repository Branch Protection Rules with Dummy PR: README Update [AAP-38413] by @art-tapin in https://github.com/ansible/metrics-utility/pull/40
* Fix failing report over data from 2024 by @MilanPospisil in https://github.com/ansible/metrics-utility/pull/39
* add pre-commit with ruff by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/42
* add data and complete test for ccspv2 file by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/41
* AAP-39154 - add pr-checks.yml by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/46
* Add missing developer documentation for the pre-commit hook setup [AAP-38768] by @art-tapin in https://github.com/ansible/metrics-utility/pull/44
* Add imported dependencies by @himdel in https://github.com/ansible/metrics-utility/pull/45
* run ruff on repo by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/50
* configure linters to use uv+ruff and remove superlinter by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/49
* AAP-39264: add helpful settings to ruff linter by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/53
* AAP-39214 - Fine tune pytest by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/55
* AAP-39220 : Run ccspv2 build_report outside awx by @himdel in https://github.com/ansible/metrics-utility/pull/51
* docs: add virtual environment handling + .gitignore update by @art-tapin in https://github.com/ansible/metrics-utility/pull/54
* Sonarcloud integration by @ShaiahWren in https://github.com/ansible/metrics-utility/pull/59
* AAP-39699 - fixing ValueError( by @larrymou9 in https://github.com/ansible/metrics-utility/pull/58
[...]

PR #122, opened 2025-04-29
Post-release - bump version to 0.5.1dev; add version mapping table

himdel:
Follows #121, #120 

Now that 0.5.0 is released, bump the version to a 0.5.1dev, just so we can tell those apart :).

And add a version mapping table to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,13 @@ Please note that this is the upstream documentation for the metrics-utility proj
 
 As the project grows, more guides and references will be added to the `/docs` folder.
 
+## Version mapping
+
+|metrics-utility version|AAP version|
+|-|-|
+|0.4.1|2.4\*|
+|0.5.0|2.5-20250507|
+
 ## Contributing
 
 Please follow our [Contributor's Guide](./docs/contributing/CONTRIBUTING.md) for details on submitting changes and documentation standards.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -170,6 +170,13 @@ Please note that this is the upstream documentation for the metrics-utility proj
 
 As the project grows, more guides and references will be added to the `/docs` folder.
 
+## Version mapping
+
+|metrics-utility version|AAP version|
+|-|-|
+|0.4.1|2.4\*|
+|0.5.0|2.5-20250507|
```
SherinV, line 178: Do we need the controller version too? 🤔

AlexSCorey (approved):
This looks good to me.

art-tapin (approved):
The version updates are consistent across all files, all tests pass successfully. 

Talking about version references, I think that build number might be omitted because it sounds like a more internal information, maybe should be double-checked with our PM.

A suggestion would be simply:
```md
|metrics-utility version|AAP version|
|-|-|
|0.4.1|2.4.x|
|0.5.0|2.5.x|
```

Other than that, LGTM 👍

README.md
himdel: Might be a good addition, yeah.

Except I still don't know how to find the controller version without guessing a commit around the right date, and looking at files to see if changes are there, so, future enhancement? :)

PR #124, opened 2025-05-12
Switch from quay.io to mirror.gcr.io

himdel:
Seems like quay login is a problem,
but these containers can be obtained from any dockerhub mirror,
=> switching to mirror.gcr.io.

(The containers are the same, except possibly more up to date versions, just different source.)
Tests still pass locally.

And added a `make compose` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 help:
-	@echo help sync test coverage lint fix
+	@echo help sync test coverage lint fix compose
 
 sync:
 	uv run sync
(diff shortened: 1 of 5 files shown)


art-tapin (approved):
Unavailability of quay.io is quite annoying, I agree :smile: 

A few scenarios I tested this (Standalone mode):

- All tests pass
- The new Makefile target works
- Scripts successfully run using the new registry
- End-to-end workflows complete successfully, i.e. gathering and report generation scenario. 

As far as I understood, "No billing data for month 2025-04" message is expected behavior when no relevant billing data exists for the period, not an issue with the container registry change.

Thanks for adding a Makefile target :1st_place_medal: 

LGTM :+1:

PR #125, opened 2025-05-13
AAP-45801: use gtime when available

ShaiahWren:
[[AAP-45801](https://issues.redhat.com/browse/AAP-45801)]

## Description

As a user of the metrics utility that generates ccspv files, I want build and gather scripts to be compatible across operating systems, including Mac and GNU. 

The change ensures that the script will use gtime if it's available (likely on macOS with GNU coreutils) and fall back to the standard time command if gtime is not found, providing better cross-platform compatibility.

### Steps to Test

Run the following commands:

./run-ccsp2-build
./run-ccsp2-gather
./run-perf
./run-renewal

### Expected Results

On all operating systems, the scripts should be executable.

### Code Quality

## Notes for Reviewers

### DIFF
--- a/run-ccsp2-build
+++ b/run-ccsp2-build
@@ -19,8 +19,7 @@ export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
 export METRICS_UTILITY_REPORT_SKU="MCT3752MO"
 export METRICS_UTILITY_REPORT_SKU_DESCRIPTION="EX: Red Hat Ansible Automation Platform, Full Support (1 Managed Node, Dedicated, Monthly)"
 
-TIME=`which gtime && echo 'gtime' || echo '/usr/bin/time'`
-
+TIME=`command -v gtime >/dev/null 2>&1 && echo 'gtime' || echo '/usr/bin/time'`
 $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./manage.py build_report --month=2024-04 --force "$@"
 
--- a/run-ccsp2-build
+++ b/run-ccsp2-build
@@ -19,7 +19,9 @@ export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
 export METRICS_UTILITY_REPORT_SKU="MCT3752MO"
 export METRICS_UTILITY_REPORT_SKU_DESCRIPTION="EX: Red Hat Ansible Automation Platform, Full Support (1 Managed Node, Dedicated, Monthly)"
 
-/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
+TIME=`which gtime && echo 'gtime' || echo 'time'`
+
+$TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./manage.py build_report --month=2024-04 --force "$@"
 
 set -x
--- a/run-ccsp2-build
+++ b/run-ccsp2-build
@@ -19,7 +19,7 @@ export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
 export METRICS_UTILITY_REPORT_SKU="MCT3752MO"
 export METRICS_UTILITY_REPORT_SKU_DESCRIPTION="EX: Red Hat Ansible Automation Platform, Full Support (1 Managed Node, Dedicated, Monthly)"
 
-TIME=`which gtime && echo 'gtime' || echo 'time'`
+TIME=`which gtime && echo 'gtime' || echo '/usr/bin/time'`
 
 $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./manage.py build_report --month=2024-04 --force "$@"
--- a/run-perf
+++ b/run-perf
@@ -20,7 +20,7 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=2000
 export OUTPUT_DATE_FROM=2026-01-01
 export OUTPUT_DATE_TO=2026-02-01
 
-TIME=`which gtime && echo 'gtime' || echo 'time'`
+TIME=`which gtime && echo 'gtime' || echo '/usr/bin/time'`
 $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./generator.py
 
@@ -36,8 +36,6 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=20000
 
 export OUTPUT_DATE_FROM=2026-02-01
 export OUTPUT_DATE_TO=2026-03-01
-
-TIME=`which gtime && echo 'gtime' || echo 'time'`
 $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./generator.py
 
@@ -55,7 +53,6 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=200000
 export OUTPUT_DATE_FROM=2026-03-01
 export OUTPUT_DATE_TO=2026-04-01
 
-TIME=`which gtime && echo 'gtime' || echo 'time'`
 $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./generator.py
 
@@ -73,7 +70,7 @@ uv run ./generator.py
 #export OUTPUT_DATE_FROM=2026-04-01
 #export OUTPUT_DATE_TO=2026-05-01
 #
-#TIME=`which gtime && echo 'gtime' || echo 'time'`
+# $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 #uv run ./generator.py
 
 
@@ -87,7 +84,6 @@ run_default_sheets() {
   export METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS
 
   echo default sheets
-  TIME=`which gtime && echo 'gtime' || echo 'time'`
   $TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
   uv run ./manage.py build_report --force "$@"
 }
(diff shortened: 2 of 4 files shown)


run-ccsp2-build
```diff
@@ -19,12 +19,14 @@ export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
 export METRICS_UTILITY_REPORT_SKU="MCT3752MO"
 export METRICS_UTILITY_REPORT_SKU_DESCRIPTION="EX: Red Hat Ansible Automation Platform, Full Support (1 Managed Node, Dedicated, Monthly)"
 
-/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
+TIME=`which gtime && echo 'gtime' || echo 'time'`
```
himdel:
```suggestion
TIME=`which gtime && echo 'gtime' || echo '/usr/bin/time'`
```

(all of them)

(because otherwise, bash will use its own builtin, which doesn't know --format either)

run-ccsp2-build
```diff
@@ -19,12 +19,14 @@ export METRICS_UTILITY_REPORT_RHN_LOGIN="test_login"
...
+$TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
 uv run ./manage.py build_report --month=2024-04 --force "$@"
 
 set -x
 ls -l metrics_utility/test/test_data/reports/2024/04/
 
 if which xlsx2csv; then
   xlsx2csv -a ./metrics_utility/test/test_data//reports/2024/04/CCSPv2-2024-04.xlsx
-fi
+fi
```
himdel: @ShaiahWren you may need to fix your editor config - it keeps removing the last newline in each file you edit.

If you're on vscode, it should be called something like "insert final newline"

run-perf
```diff
@@ -36,7 +37,8 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=20000
 export OUTPUT_DATE_FROM=2026-02-01
 export OUTPUT_DATE_TO=2026-03-01
 
-/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
+TIME=`which gtime && echo 'gtime' || echo 'time'`
```
himdel: you only need to do this once per file - the variable stays set throughout the rest of the file so no need to set it multiple times

run-ccsp2-build
ShaiahWren: Good catch

run-perf
```diff
@@ -36,7 +37,6 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=20000
 export OUTPUT_DATE_FROM=2026-02-01
 export OUTPUT_DATE_TO=2026-03-01
 
-/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
```
himdel, line 39: removed too much?

These should still be here, just, using `$TIME`

ShaiahWren: Thanks, corrected :) and do you think I should change anything in the dev setup at this point? or does the updated script speak for itself

himdel: Looking at the diff in https://github.com/ansible/metrics-utility/pull/125/files#diff-131add6a5e2097d535cac484706b46ea3a5d32c9697485fa1161d4c4173086b1 ..

I still see a couple of removed calls

But otherwise good I think :)

run-perf
```diff
@@ -70,7 +70,7 @@ uv run ./generator.py
 #export OUTPUT_DATE_FROM=2026-04-01
 #export OUTPUT_DATE_TO=2026-05-01
 #
-#/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
+$TIME --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n' \
```
himdel: this part was commented out but now isn't

art-tapin (commented):
Hi @ShaiahWren! I just saw that @himdel already pointed out the issue in `run-perf` where a previously commented part is now uncommented.

A couple other friendly suggestions to make your PR even better:

1. The `which` command could be improved with error redirection:
   ```bash
   TIME=`which gtime 2>/dev/null && echo 'gtime' || echo '/usr/bin/time'`
   ```
   This prevents error messages when gtime isn't found.

2. For maximum portability, you might consider using `command -v` instead of `which`:
   ```bash
   TIME=`command -v gtime >/dev/null 2>&1 && echo 'gtime' || echo '/usr/bin/time'`
   ```
   The `command` builtin is available in most POSIX shells, while `which` might not be.

3. Placing the `TIME` declaration higher up in the scripts would be a nice readability improvement (in run-ccsp2-build, run-ccsp2-gather, and run-renewal).

4. And don't forget about the newlines at the end of files that @himdel mentioned!

Great work! :hugs:

AlexSCorey:
I'm not sure this is working for me.  
when I run those files on my machine I get 
```acorey  >  metrics                                                               10:13:16
(metrics-utility) metrics-utility  ((dd03b563)) >  ./run-ccsp2-build             10:13:21

/usr/bin/time: illegal option -- -
usage: time [-al] [-h | -p] [-o file] utility [argument ...]
(metrics-utility) metrics-utility  ((dd03b563)) >  ./run-ccsp2-gather            10:13:28

ccsp2-gather-latest
localhost:5432 - accepting connections
localhost:5432 - accepting connections
/usr/bin/time: illegal option -- -
usage: time [-al] [-h | -p] [-o file] utility [argument ...]
(metrics-utility) metrics-utility  ((dd03b563)) >  ./run-perf                    10:13:44

gen 10,000
/usr/bin/time: illegal option -- -
usage: time [-al] [-h | -p] [-o file] utility [argument ...]
(metrics-utility) metrics-utility  ((dd03b563)) >  ./run-renewal                 10:13:59


renewal-latest
localhost:5432 - accepting connections
localhost:5432 - accepting connections
/usr/bin/time: illegal option -- -
usage: time [-al] [-h | -p] [-o file] utility [argument ...]
```
and when I run them inside the metrics-utility docker container 
```# ./run-ccsp2-gather
ccsp2-gather-latest
./run-ccsp2-gather: 29: pg_isready: not found
./run-ccsp2-gather: 30: podman: not found
# ./run-ccsp2-build
./run-ccsp2-build: 23: /usr/bin/time: not found
# uv sync
Resolved 52 packages in 5ms
Audited 49 packages in 1ms
# ./run-perf
gen 10,000
./run-perf: 24: /usr/bin/time: not found
# ./run-renewal
renewal-latest
./run-renewal: 16: pg_isready: not found
./run-renewal: 17: podman: not found
```

himdel:
@AlexSCorey you may need to `brew install coreutils`.

I guess we *could* change the logic to something like 

```sh
if [ `uname -s` = 'Linux' ]; then
  TIME=/usr/bin/time
else
  if command -v gtime; then
    TIME=gtime
  else
    echo "Please install coreutils" 1>&2
    exit 1
  fi
fi
```

but .. since this is really just a "quick how do I run this thing" script for us, maybe not needed?

(And yeah, not expected to work inside the container, too many things missing there.)

EDIT: or, feature detection + simpler fallback like ``TIME=`env time --format="" true >/dev/null 2>/dev/null && echo "/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n'" || echo time` ``? (+-, may need doubling backslashes),
and then replace all those `$TIME --format...` with just `$TIME`?

ShaiahWren:
@himdel / @AlexSCorey , how about
TIME=`env time --format="" true >/dev/null 2>/dev/null && echo "/usr/bin/time --format='\n\nmemory %MK\ncpu %P\ntime %es\n\n'" || echo time`
$TIME

This doesn't prioritize gtime, but, it's available for those who need it. :)

run-perf
```diff
@@ -53,7 +53,7 @@ export MAIN_JOBEVENT_UNIQUE_SIZE=200000
 export OUTPUT_DATE_FROM=2026-03-01
 export OUTPUT_DATE_TO=2026-04-01
```
AlexSCorey, line 54: Why do we have output date for next year?  I think this is causing a failure, and we should file a bug, but it is probably not within the scope of this pr.

```Traceback (most recent call last):
  File "/Users/acorey/Ansible/metrics-utility/./generator.py", line 407, in <module>
    main.load()
  File "/Users/acorey/Ansible/metrics-utility/./generator.py", line 289, in load
    data = process_tarballs(file, temp_dir, enabled_set=self.selected)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/acorey/Ansible/metrics-utility/./generator.py", line 164, in process_tarballs
    return ProcessTarballs().process_tarballs(path, temp_dir)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/acorey/Ansible/metrics-utility/metrics_utility/automation_controller_billing/extract/base.py", line 62, in process_tarballs
    config = self.load_config(os.path.join(temp_dir, 'config.json'))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/acorey/Ansible/metrics-utility/./generator.py", line 157, in load_config
    with open(file_path) as f:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/n8/_72dbnrs6nq9stlk928sy15h0000gn/T/metrics-generator-loady5wlfibo/config.json'```

AlexSCorey (approved):
I was able to pull this down and get it to work.  All looks good.  I think there is a bug unrelated to Shaiah's work.  @ShaiahWren agreed to file the bug in jira for us.  Thanks Shaiah.

run-perf
himdel: Hm, that looks like you might have a tarball that's missing a `config.json` inside? Do you know which file? Is it supposed to be valid or did we mix in some of those tarballs for testing how things break?

(Future dates were intentional, so as not to mix the generated data with anything real .. but maybe we'll want to move them to the past, to future-proof things :))

himdel: Is it `metrics_utility/test/test_data/data/2025/03/01/9bf6f6c3-eeb3-4523-8e8d-80f3d4d01978-2025-02-25-000000+0000-2025-02-25-000000+0000-0.tar.gz` (added in #118)

If so, we just need to move those tarballs outside the range that generator uses (or adjust the range to something that works)

((Or, add an except that just returns an empty dict, but... the issue is in the data.))

himdel (approved):
Confirmed working on linux (+- the unrelated run-perf issue)

PR #112, opened 2025-04-09
Remove dead code

himdel:
Remove dead code as reported by vulture.
(Manually filtered to keep code that IS used.)

(TODO for later: add a workflow running vulture and a whitelist for known good places)

PR #128, opened 2025-05-21
AAP-45505 - Asserts that exceptions for build and gather appear with the proper messaging.

AlexSCorey:
AAP-45505
## Description

I've moved some validation helper functions into commands/validate, and I've added some tests for buld_report to ensure that proper validation occurs for that command.  Next I'll move to the gather commands.

### Steps to Test

1. You can run our existing test suite. The new test file should pass as well.

### DIFF
--- a/metrics_utility/automation_controller_billing/helpers.py
+++ b/metrics_utility/automation_controller_billing/helpers.py
@@ -9,7 +9,10 @@ import pandas as pd
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 
-from metrics_utility.exceptions import UnparsableParameter
+from metrics_utility.exceptions import DateFormatError, UnparsableParameter
+
+
+DAY_OR_MONTH_REGEX_PATTERN = r'.^\d+(d|day|days|m|mo|month|months)$'
 
 
 def parse_json_array(x):
@@ -113,3 +116,22 @@ def parse_number_of_days(date_option):
         raise UnparsableParameter(f"Can't parse parameter value {date_option}")
 
     return days
+
+
+def handle_month(month):
+    """Process month argument"""
+    if month is not None:
+        try:
+            date = datetime.datetime.strptime(f'{month}', '%Y-%m')
+        except ValueError:
+            raise DateFormatError('Invalid --month format. Supported date format: YYYY-MM')
+    else:
+        """Return last month if no month was passed"""
+        beginning_of_the_month = datetime.datetime.today().replace(day=1)
+        beginning_of_the_previous_month = beginning_of_the_month - relativedelta(months=1)
+        date = beginning_of_the_previous_month
+        y = date.strftime('%Y')
+        m = date.strftime('%m')
+        month = f'{y}-{m}'
+
+    return month, date, date + relativedelta(months=1)
--- a/metrics_utility/automation_controller_billing/helpers.py
+++ b/metrics_utility/automation_controller_billing/helpers.py
@@ -12,7 +12,9 @@ from dateutil.relativedelta import relativedelta
 from metrics_utility.exceptions import DateFormatError, UnparsableParameter
 
 
-DAY_OR_MONTH_REGEX_PATTERN = r'.^\d+(d|day|days|m|mo|month|months)$'
+ALLOWED_EPHEMERAL_PATTERN = r'^\d+(d|day|days|m|mo|month|months)$'
+SINCE_AND_UNTIL_GATHER_PATTERN = r'^\d+[dm]$|^\d{4}-\d{2}-\d{2}$'
+SINCE_AND_UNTIL_BUILD_PATTERN = r'^\d+(d|mo|month|months|m)$|^\d{4}-\d{2}-\d{2}$'
 
 
 def parse_json_array(x):
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -4,7 +4,6 @@ import os
 
 from datetime import timezone
 
-from dateutil import parser
 from django.core.management.base import BaseCommand
 
 from metrics_utility.automation_controller_billing.dataframe_engine.factory import Factory as DataframeEngineFactory
@@ -12,7 +11,6 @@ from metrics_utility.automation_controller_billing.extract.factory import Factor
 from metrics_utility.automation_controller_billing.helpers import (
     handle_month,
     parse_date_param,
-    parse_number_of_days,
 )
 from metrics_utility.automation_controller_billing.report.factory import Factory as ReportFactory
 from metrics_utility.automation_controller_billing.report_saver.factory import Factory as ReportSaverFactory
@@ -43,36 +41,35 @@ class Command(BaseCommand):
     help = 'Gather Automation Controller billing data'
 
     def __init__(self):
-        # help = 'Gather Automation Controller billing data'
-        self.help_since = (
-            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
-        )
-        self.help_until = (
-            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
-        )
-        self.help_time_frame_extra_params = (
-            'Missing required parameter --month, --until, or --since. Metrics utility requires a value for at least \n'
-            'one of the following: month, since, until.'
-        )
-        self.help_month = 'Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report will \n'
-        "be generated if it doesn't exists already."
-        self.help_ephemeral = 'Duration in months or days to determine if host is ephemeral. Months are taken as 30days duration.\n'
-        'Example: --ephemeral=3months, or --ephemeral=3days'
+        super().__init__()
+
+        self.help = {
+            'since': """Start date for collection including (e.g. --since=2023-12-20), a number of minutes ago (--until=2m),
+              a number of days ago (--since=5d), or a number of months (--since=2m).""",
+            'until': 'End date for collection including (e.g. --until=2023-12-21), a number of minutes (--until=2m), '
+            'a number of days ago (--until=5d), or a number of months (--until=2m).',
+            'time_frame_extra_params': 'Missing required parameter --month, --until, or --since. Metrics utility requires a value for at least '
+            'one of the following: month, since, until.',
+            'month': """Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report
+             will be generated if it doesn't exists already.""",
+            'ephemeral': """Duration in months or days to determine if host is ephemeral. Months are taken as 30days duration.
+            Example: --ephemeral=3months, or --ephemeral=3days""",
+        }
 
     def add_arguments(self, parser):
-        parser.add_argument('--month', dest='month', action='store', help=self.help_month)
-        parser.add_argument('--since', dest='since', action='store', help=self.help_since)
+        parser.add_argument('--month', dest='month', action='store', help=self.help.get('month'))
+        parser.add_argument('--since', dest='since', action='store', help=self.help.get('since'))
         parser.add_argument(
             '--until',
             dest='until',
             action='store',
-            help=self.help_until,
+            help=self.help.get('until'),
         )
         parser.add_argument(
             '--ephemeral',
             dest='ephemeral',
             action='store',
-            help=self.help_ephemeral,
+            help=self.help.get('ephemeral'),
         )
         parser.add_argument(
             '--force',
@@ -96,16 +93,15 @@ class Command(BaseCommand):
         options,
     ):
         param = options.get(param_name)
-        if (options.get('month') is not None or param is None) and param_name != 'ephemeral':
+        if options.get('month') is not None or param is None:
             return None
-        if param:
-            return parse_date_param(param) if parser.parse(param) else parse_number_of_days(param)
+        return parse_date_param(param)
 
     def _handle(self, *args, **options):
         self.init_logging()
         og_month, month, next_month = handle_month(options.get('month') or None)
 
-        validate_build_extra_params(self, options)
+        validate_build_extra_params(self.help, options)
         opt_month = og_month if options.get('month') else None
         opt_until = self._parse_param('until', options)
         opt_since = self._parse_param('since', options)
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -4,17 +4,34 @@ import os
 
 from datetime import timezone
 
-from dateutil.parser import parse as date_parse
-from dateutil.relativedelta import relativedelta
+from dateutil import parser
 from django.core.management.base import BaseCommand
 
 from metrics_utility.automation_controller_billing.dataframe_engine.factory import Factory as DataframeEngineFactory
 from metrics_utility.automation_controller_billing.extract.factory import Factory as ExtractorFactory
-from metrics_utility.automation_controller_billing.helpers import parse_date_param
+from metrics_utility.automation_controller_billing.helpers import (
+    handle_month,
+    parse_date_param,
+    parse_number_of_days,
+)
 from metrics_utility.automation_controller_billing.report.factory import Factory as ReportFactory
 from metrics_utility.automation_controller_billing.report_saver.factory import Factory as ReportSaverFactory
-from metrics_utility.exceptions import BadRequiredEnvVar, BadShipTarget, MissingRequiredEnvVar
-from metrics_utility.management.validation import handle_directory_ship_target, handle_not_crc, handle_not_s3, handle_s3_ship_target
+from metrics_utility.exceptions import (
+    BadParameter,
+    BadRequiredEnvVar,
+    BadShipTarget,
+    DateFormatError,
+    MissingRequiredEnvVar,
+    MissingRequiredParameter,
+    UnparsableParameter,
+)
+from metrics_utility.management.validation import (
+    handle_directory_ship_target,
+    handle_not_crc,
+    handle_not_s3,
+    handle_s3_ship_target,
+    validate_build_extra_params,
+)
 from metrics_utility.metric_utils import get_optional_collectors
 
 
@@ -25,27 +42,37 @@ class Command(BaseCommand):
 
     help = 'Gather Automation Controller billing data'
 
-    def add_arguments(self, parser):
-        parser.add_argument(
-            '--month',
-            dest='month',
-            action='store',
-            help='Month the report will be generated for, with format YYYY-MM. '
-            "If this params isn't provided, previou month report will be"
-            "generated if it doesn't exists already.",
+    def __init__(self):
+        # help = 'Gather Automation Controller billing data'
+        self.help_since = (
+            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
         )
-        parser.add_argument('--since', dest='since', action='store', help='Date or number of days/months ago we want to generate the reports for.')
+        self.help_until = (
+            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+        )
+        self.help_time_frame_extra_params = (
+            'Missing required parameter --month, --until, or --since. Metrics utility requires a value for at least \n'
+            'one of the following: month, since, until.'
+        )
+        self.help_month = 'Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report will \n'
+        "be generated if it doesn't exists already."
+        self.help_ephemeral = 'Duration in months or days to determine if host is ephemeral. Months are taken as 30days duration.\n'
+        'Example: --ephemeral=3months, or --ephemeral=3days'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--month', dest='month', action='store', help=self.help_month)
+        parser.add_argument('--since', dest='since', action='store', help=self.help_since)
         parser.add_argument(
             '--until',
             dest='until',
             action='store',
-            help='Date or number of days/months ago we want to generate the reports until. Not available for renewal guidance report',
+            help=self.help_until,
         )
         parser.add_argument(
             '--ephemeral',
             dest='ephemeral',
             action='store',
-            help='Duration in months or days to determine if host is ephemeral. Months are takenas 30days duration.',
+            help=self.help_ephemeral,
         )
         parser.add_argument(
             '--force',
@@ -61,43 +88,34 @@ class Command(BaseCommand):
         handler.setLevel(logging.DEBUG)
         handler.setFormatter(logging.Formatter('%(message)s'))
         self.logger.addHandler(handler)
-        self.logger.propagate = False
+        self.logger.propagate = True
 
-    def handle(self, *args, **options):
+    def _parse_param(
+        self,
+        param_name,
+        options,
+    ):
+        param = options.get(param_name)
+        if (options.get('month') is not None or param is None) and param_name != 'ephemeral':
+            return None
+        if param:
+            return parse_date_param(param) if parser.parse(param) else parse_number_of_days(param)
+
+    def _handle(self, *args, **options):
         self.init_logging()
+        og_month, month, next_month = handle_month(options.get('month') or None)
 
-        # parse params
-        opt_month = options.get('month') or None
-        opt_month, month, next_month = self._handle_month(opt_month)
-
-        # Since and ephemenral params are specific to subset of reports
-        opt_since = None
-        opt_ephemeral = None
-
-        opt_since = options.get('since') or None
-        opt_since = parse_date_param(opt_since)
-
-        opt_ephemeral = options.get('ephemeral') or None
-
-        opt_until = None  # Initialize opt_until to None
-
-        if opt_since is not None:
-            # If --since was provided, then process --until
-            user_provided_until = options.get('until')
-            if user_provided_until is None:
-                # default --until to today
-                opt_until = datetime.datetime.now(timezone.utc).replace(hour=0, minute=0, second=0, microsecond=0)
-                self.logger.info(f'--until parameter not provided with --since, defaulting to: {opt_until.date()}')
-            else:
-                # If --since and --until were both provided, parse the user's --until
-                opt_until = parse_date_param(user_provided_until)
-
+        validate_build_extra_params(self, options)
+        opt_month = og_month if options.get('month') else None
+        opt_until = self._parse_param('until', options)
+        opt_since = self._parse_param('since', options)
         opt_force = options.get('force')
+
         ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET', None)
         extra_params = self._handle_extra_params(ship_target)
         extra_params['opt_since'] = opt_since
         extra_params['opt_until'] = opt_until
-        extra_params['opt_ephemeral'] = opt_ephemeral
+        extra_params['opt_ephemeral'] = options.get('ephemeral') or None
         extra_params['month_since'] = month
         extra_params['month_until'] = next_month
 
@@ -152,6 +170,24 @@ class Command(BaseCommand):
         # Save the report to the configured destination
         report_saver_engine.save(report_spreadsheet)
         self.logger.info(f'Report generated into {ship_target}: {report_saver_engine.report_spreadsheet_destination_path}')
+
+    def handle(self, *args, **options):
+        try:
+            self._handle(*args, **options)
+        except (
+            BadShipTarget,
+            MissingRequiredEnvVar,
+            BadRequiredEnvVar,
+            MissingRequiredParameter,
+            UnparsableParameter,
+            BadParameter,
+            DateFormatError,
+        ) as e:
+            self.logger.error(str(e))
+            exit(1)
+        except Exception as e:
+            self.logger.exception(e)
+            exit(1)
 
     def _handle_ship_target(self, ship_target):
         if ship_target in ['controller_db', 'directory']:
@@ -210,18 +246,3 @@ class Command(BaseCommand):
             }
         )
         return base
-
-    def _handle_month(self, month):
-        # Process month argument
-        if month is not None:
-            date = date_parse(f'{month}-01')
-        else:
-            # Return last month if no month was passed
-            beginning_of_the_month = datetime.date.today().replace(day=1)
-            beginning_of_the_previous_month = beginning_of_the_month - relativedelta(months=1)
-            date = beginning_of_the_previous_month
-            y = date.strftime('%Y')
-            m = date.strftime('%m')
-            month = f'{y}-{m}'
-
-        return month, date, date + relativedelta(months=1)
--- a/metrics_utility/management/commands/gather_automation_controller_billing_data.py
+++ b/metrics_utility/management/commands/gather_automation_controller_billing_data.py
@@ -25,25 +25,25 @@ from metrics_utility.management.validation import (
 )
 
 
-help_since = 'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
-help_until = 'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
-
-
 class Command(BaseCommand):
     """
     Gather Automation Controller billing data
     """
 
     def __init__(self):
-        self.help_since = (
-            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
-        )
-        self.help_until = (
-            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
-        )
-        self.help_time_frame_extra_params = (
-            'Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since, until.'
-        )
+        super().__init__()
+        self.help = {
+            'since': (
+                'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), '
+                'or a number of months (--since=2m).'
+            ),
+            'until': (
+                'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+            ),
+            'time_frame_extra_params': (
+                'Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since, until.'
+            ),
+        }
 
     def add_arguments(self, parser):
         parser.add_argument('--dry-run', dest='dry-run', action='store_true', help='Gather billing metrics without shipping.')
@@ -53,13 +53,13 @@ class Command(BaseCommand):
             '--since',
             dest='since',
             action='store',
-            help=help_since,
+            help=self.help.get('since'),
         )
         parser.add_argument(
             '--until',
             dest='until',
             action='store',
-            help=help_until,
+            help=self.help.get('until'),
         )
 
     def init_logging(self):
@@ -84,7 +84,8 @@ class Command(BaseCommand):
     def _handle(self, *args, **options):
         self.init_logging()
 
-        handle_validate_date_param(self, options)
+        handle_validate_date_param(options.get('since', None), self.help.get('since'), 'gather')
+        handle_validate_date_param(options.get('until', None), self.help.get('until'), 'gather')
 
         opt_ship = options.get('ship')
         opt_dry_run = options.get('dry-run')
@@ -150,8 +151,8 @@ class Command(BaseCommand):
 
     def _handle_interval(self, opt_since, opt_until):
         # Process since argument
-        since = self._handle_datelike(opt_since, help=help_since)
+        since = self._handle_datelike(opt_since, help=self.help.get('since'))
 
         # Process until argument
-        until = self._handle_datelike(opt_until, help=help_until)
+        until = self._handle_datelike(opt_until, help=self.help.get('until'))
         return since, until
--- a/metrics_utility/management/commands/gather_automation_controller_billing_data.py
+++ b/metrics_utility/management/commands/gather_automation_controller_billing_data.py
@@ -21,6 +21,7 @@ from metrics_utility.management.validation import (
     handle_not_crc,
     handle_not_s3,
     handle_s3_ship_target,
+    handle_validate_date_param,
 )
 
 
@@ -33,7 +34,16 @@ class Command(BaseCommand):
     Gather Automation Controller billing data
     """
 
-    help = 'Gather Automation Controller billing data'
+    def __init__(self):
+        self.help_since = (
+            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
+        )
+        self.help_until = (
+            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+        )
+        self.help_time_frame_extra_params = (
+            'Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since, until.'
+        )
 
     def add_arguments(self, parser):
         parser.add_argument('--dry-run', dest='dry-run', action='store_true', help='Gather billing metrics without shipping.')
@@ -64,7 +74,7 @@ class Command(BaseCommand):
         try:
             self._handle(self, *args, **options)
             exit(0)
-        except (BadShipTarget, MissingRequiredEnvVar, BadRequiredEnvVar, FailedToUploadPayload) as e:
+        except (BadShipTarget, MissingRequiredEnvVar, BadRequiredEnvVar, FailedToUploadPayload, UnparsableParameter) as e:
             self.logger.error(e.name)
             exit(1)
         except Exception as e:
@@ -73,6 +83,9 @@ class Command(BaseCommand):
 
     def _handle(self, *args, **options):
         self.init_logging()
+
+        handle_validate_date_param(self, options)
+
         opt_ship = options.get('ship')
         opt_dry_run = options.get('dry-run')
         opt_since = options.get('since') or None
@@ -119,13 +132,7 @@ class Command(BaseCommand):
     def _handle_datelike(self, value, help=''):
         if not value:
             return None
-
-        if value.isdigit():
-            raise UnparsableParameter(f'Bare numbers are not valid ({help})')
-
-        # Process ret argument
-        ret = None
-
+        # # Process ret argument
         if value.endswith('d'):
             days_ago = int(value[0:-1])
             ret = (datetime.datetime.now() - datetime.timedelta(days=days_ago - 1)).replace(hour=0, minute=0, second=0, microsecond=0)
@@ -147,5 +154,4 @@ class Command(BaseCommand):
 
         # Process until argument
         until = self._handle_datelike(opt_until, help=help_until)
-
         return since, until
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -1,7 +1,11 @@
 import logging
 import os
+import re
 
-from metrics_utility.exceptions import MissingRequiredEnvVar
+from dateutil import parser
+
+from metrics_utility.automation_controller_billing.helpers import DAY_OR_MONTH_REGEX_PATTERN, parse_date_param
+from metrics_utility.exceptions import BadParameter, MissingRequiredEnvVar, MissingRequiredParameter, UnparsableParameter
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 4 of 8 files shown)


metrics_utility/test/util.py
```diff
@@ -66,4 +66,4 @@ def run_gather_int(env, options):
 
     with temporary_env(env):
         # .handle does exit(0), breaking tests
-        Command()._handle(**options)
+        Command().handle(**options)
```
himdel: this needs to stay as it was, the tests can handle exceptions, but not exit

metrics_utility/management/commands/build_report.py
```diff
@@ -61,23 +67,18 @@ def init_logging(self):
...
+        opt_month, month, next_month, opt_since, opt_until, opt_ephemeral = validate_date_extra_params(options)
 
-        # parse params
-        opt_month = options.get('month') or None
-        opt_month, month, next_month = self._handle_month(opt_month)
-
-        # Since and ephemenral params are specific to subset of reports
-        opt_since = None
-        opt_ephemeral = None
+        if opt_month is None:
```
himdel: surprising .. if validate_* handles setting opt_month, why doesn't it handle setting opt_since and opt_until?

metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -109,8 +110,22 @@ def _handle_datelike(self, value, help=''):
         if not value:
             return None
 
-        if value.isdigit():
-            raise UnparsableParameter(f'Bare numbers are not valid ({help})')
+        if (
+            isinstance(value, str)
+            and (not value.endswith('m') and not value.endswith('d'))
+            and not isinstance(value, datetime.datetime)
```
himdel: if you're getting a datetime here, somethings is getting called twice, or the tests are passing in datetimes instead of strings (and thus not testing real values)

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def test_invalid_month_format(caplog):
+    bad_inputs = ['abc', '12', 'mo3', '3mon', '3months', '3m', '3mo']
+
+    for bad_input in bad_inputs:
+        handle_exception(caplog, env_vars, {'month': bad_input}, 'Invalid --month format. Supported date format: YYYY-MM')
+
+
+def handle_exception(caplog, env_vars, params, error_message):
```
himdel: this will not be needed if the test just calls _handle, and tests the right exception was returned

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,113 @@
...
+    'METRICS_UTILITY_SHIP_TARGET': 'directory',
+}
+
+
+def handle_gather_exception(env_vars, params, error_message):
+    with pytest.raises(Exception) as exc_info:
+        run_gather_int(env_vars, params)
+
+        found_error = any(error_message in record.message for record in exc_info.value)
+        assert found_error, f'Expected error message not found in logs: {exc_info.value}'
```
himdel: SonarQube was right here ("Don’t perform an assertion here; An exception is expected to be raised before its execution."), there are 0 asserts in this test... (well, except for `pytest.raises` itself, so, it **is** checking there was an Exception, just not the rest)

`run_gather_int` raises an exception, as expected, ending that whole `with` block
python then continues after the `with`  .. but there's no code left there

```suggestion
    with pytest.raises(Exception) as exc_info:
        run_gather_int(env_vars, params)

    found_error = any(error_message in record.message for record in exc_info.value)
    assert found_error, f'Expected error message not found in logs: {exc_info.value}'
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def handle_build_exception(env_vars, params, error_message):
+    with pytest.raises(SystemExit) as exc_info:
+        run_build_int(env_vars, params)
+
+        # Optionally check exit code
+        assert exc_info.value.code == 1
+        found_error = any(error_message in record.message for record in exc_info.value)
+        assert found_error, f'Expected error message not found in logs: {exc_info.value}'
```
himdel:
```suggestion
    with pytest.raises(SystemExit) as exc_info:
        run_build_int(env_vars, params)

    # Optionally check exit code
    assert exc_info.value.code == 1
    found_error = any(error_message in record.message for record in exc_info.value)
    assert found_error, f'Expected error message not found in logs: {exc_info.value}'
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,113 @@
...
+        workbook = load_workbook(filename=month_file_path)
+        worksheet = read_excel(month_file_path, sheet_name='Usage Reporting')
+        cell_value = worksheet.iat[3, 1]
+        assert cell_value == '2025-04', f"Expected '2025-04' but got '{cell_value}'"
+    finally:
+        workbook.close()
+        os.remove(month_file_path)
+
+
+def test_renewal_guidancev2_fails_with_until_params(caplog):
```
himdel: oh, sorry, most of renewal guidance v2 was removed in #112 (specifically https://github.com/ansible/metrics-utility/pull/112/files#diff-61351d26627432d40286d834a318626e8f4baff10049dc819172acb4aae2caaf)

what remained was 2 factories still trying to do something useful with it, and ending up with renewal guidance v1 but with the wrong dataframe(s)

removing in #131 

so .. probably no tests for `RENEWAL_GUIDANCEv2` either? (and helptexts)

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,113 @@
...
+    bad_inputs = ['2', '2y', 'mo3', '3weeks', '3w']
+    args = ['until', 'since']
+    for bad_input in bad_inputs:
+        for arg in args:
+            help_text = GatherBase.help_since if arg == 'since' else GatherBase.help_until
+            handle_gather_exception(gather_env_vars, {arg: bad_input}, help_text)
+
+
+def test_required_gather_params():
+    error_message = f"""Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since,
```
himdel: Does it?

https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/automation_controller_billing/collector.py#L51-L52
says
> since: (datetime) - low threshold of data changes (max. and default - 4 weeks ago)
> until: (datetime) - high threshold of data changes (defaults to now)

and the code in `_calculate_collection_interval` seems to match, so I think these are the defaults for gather (`Collector` `gather` gets called with since/until` or None` from the command, that calls `_gather_initialize`, and that calls `_calculate_collection_interval(since, until)`)

We may also want to add a check that `since` is not more then 4 weeks in the past (and neither is `until`, but that's probably already satisfied by the since < until check :)). But then also that check doesn't take into account that default value, since it only defaults to that later on... so maybe for gather, the validator should call `_calculate_collection_interval` before doing checks that are not already handled by that .. and we can shift the fatal warnings text inside those exceptions :) .. hm, except you'd probably have to split out the `last_gather` bits before calling that method from the validator, since those do need to connect to the db, which will not be possible during param parsing once we move the connection settings to params :) .. but that can probably wait for then too, iff it works :)

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +79,37 @@ def init_logging(self):
...
         self.logger.addHandler(handler)
-        self.logger.propagate = False
+        self.logger.propagate = True
+
+    def _parse_param(
+        self,
+        param,
+        options,
+    ):
+        if options.get('month') is not None:
```
himdel: this doesn't work when you're calling `_parse_param` with `'ephemeral'` too

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +79,37 @@ def init_logging(self):
...
+        opt_until = self._parse_param('until', options)
+        opt_since = self._parse_param('since', options)
 
         opt_force = options.get('force')
         ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET', None)
         extra_params = self._handle_extra_params(ship_target)
         extra_params['opt_since'] = opt_since
         extra_params['opt_until'] = opt_until
-        extra_params['opt_ephemeral'] = opt_ephemeral
+        extra_params['opt_ephemeral'] = self._parse_param('ephemeral', options)
```
himdel: opt_ephemeral is only used with `parse_number_of_days` .. allows only `<number>d,day,days,mo,month,months`

but `_parse_param` calls `_parse_date_param` , which allows `<number>d,mo,month,months`, close enought .. but also `m` for minutes, and actual dates

so, probably best to use a separate validator function? We could probably just call `parse_number_of_days` on it, unless there are more cases to handle, maybe even so.

metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -108,13 +125,7 @@ def _handle_ship_target(self, ship_target):
     def _handle_datelike(self, value, help=''):
         if not value:
             return None
-
-        if value.isdigit():
-            raise UnparsableParameter(f'Bare numbers are not valid ({help})')
```
himdel, line 124: This check was necessary, people were calling it with bare numbers and getting weird behaviour, because `parser.parse(2000)` just gives you a `datetime` that's today at midnight, but in the year 2000, no timezone, while `parser.parse(20)` gives you the 20th of the current month, etc.

(And I'm not seeing it moved anywhere else?)

metrics_utility/management/validation.py
```diff
@@ -58,3 +62,47 @@ def handle_crc_ship_target(ship_target):
...
+
+def validate_date_day_month_pattern(date: str, help_text):
+    date_pattern = r'^\d{4}-\d{2}(-\d{2})?$'
+    day_month_pattern = r'.*[dm]$'
+    if date is not None and not (re.match(date_pattern, date) or re.match(day_month_pattern, date)):
+        raise UnparsableParameter(help_text)
+
+
+def validate_date_extra_params(self, options):
+    if 'gather_automation_controller_billing_data' in self.__module__:
```
himdel: why not just call the right function instead of this one?

(I'm not sure it's ever a good idea to use `self.__module__`, maybe outside some test mocks?)

metrics_utility/management/validation.py
```diff
@@ -58,3 +62,47 @@ def handle_crc_ship_target(ship_target):
...
+def validate_gather_timeframe_params(self, options):
+    since = options.get('since') or None
+    until = options.get('until') or None
+
+    if since is None and until is None:
+        raise MissingRequiredParameter(
+            f"""Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since, until. \n\n
+            Since help text: {self.help_since} Until help text: {self.help_until}"""
+        )
+    args = (since, self.help_since) if since is None else (until, self.help_until)
```
himdel: does this mean we only validate `since` if it is none, and if `since` is present, we only validate `until`?

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +79,37 @@ def init_logging(self):
...
-        opt_since = options.get('since') or None
-        opt_since = parse_date_param(opt_since)
-
-        opt_ephemeral = options.get('ephemeral') or None
-
-        opt_until = None  # Initialize opt_until to None
-
-        if opt_since is not None:
-            # If --since was provided, then process --until
-            user_provided_until = options.get('until')
```
himdel, line 86: previously, `opt_until` would not be set (or parsed) unless `opt_since` is as well

but now we're setting both when the param is present .. which changes `_since_until` in ReportCCSPv2 to prefer `until` over `month` .. and `validate_date_extra_params` only warns about that case, but doesn't error out.

metrics_utility/test/test_extra_params_validation.py
himdel: > `assert exc_info.value.code == 1`

there's no exit code with `run_build_int`, just the exception

metrics_utility/management/commands/gather_automation_controller_billing_data.py
AlexSCorey: So before we get to this function we validate the values of of since and until to ensure they match.  That validation begins [here](https://github.com/ansible/metrics-utility/pull/128/files/61fd7301933dd18b996a71f2796d291ba8f1006a#diff-bb98f7ff0b0f4e4c66ebfeb4296c00cf73d8e86eafcb1fd12ba5c7eca4aeb2a7R85).  Bare numbers are no longer allow and will throw a validation error.

metrics_utility/management/validation.py
AlexSCorey: Yeah, it does seem that way.  I'll fix it.

metrics_utility/management/commands/gather_automation_controller_billing_data.py
himdel: Validation there is done by calling `parse_date_param`.

But that one does accept bare numbers, precisely because it ends up in the `parser.parse` branch, which accepts those.

metrics_utility/test/test_extra_params_validation.py
AlexSCorey: I'm not real sure how to proceed with this.  Should I just remove the test?

metrics_utility/management/commands/build_report.py
AlexSCorey: Do we want it to error out? I thought we had a flow that just ignore until and since if month was set?  I added the warning just so that the user would know what happened if they got unexpected results.

metrics_utility/test/test_extra_params_validation.py
himdel: Well, sounds like it's not just the test that doesn't match what the code does, it's also the check that this test is testing.

If the code previously worked with no since/until for gather, it should keep working.
If it had these defaults, those defaults need to be taken into account during the validation - once testing for required variables is done, we need to use those default values so that the checks after it do what they claim to do (comparing since < until is only useful if you already have the right value, including the default if not given).

metrics_utility/management/commands/build_report.py
himdel: It's probably ok to make it an error, yeah.

But if we don't make it an error, we need to keep `until` unset when month is present to not break that logic.

AlexSCorey: I'll make it error out.

metrics_utility/automation_controller_billing/helpers.py
```diff
@@ -9,7 +9,11 @@
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 
-from metrics_utility.exceptions import UnparsableParameter
+from metrics_utility.exceptions import DateFormatError, UnparsableParameter
+
+
+DATE_REGEX_PATTERN = r'^\d{4}-\d{2}(-\d{2})?$'
+DAY_OR_MONTH_REGEX_PATTERN = r'.*(d|day|days|m|mo|month|months)$'
```
himdel:
```suggestion
DAY_OR_MONTH_REGEX_PATTERN = r'^\d+(d|day|days|m|mo|month|months)$'
```

(otherwise "Thursday" would be a valid value too)

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +85,34 @@ def init_logging(self):
...
+    def _parse_param(
+        self,
+        param_name,
+        options,
+    ):
+        param = options.get(param_name)
+        if (options.get('month') is not None or param is None) and param_name != 'ephemeral':
+            return None
+        if param:
+            return parse_date_param(param) if re.match(DATE_REGEX_PATTERN, param) else parse_number_of_days(param)
```
himdel: I think we may need to simplify this function, it may be doing too much...

Right now, it will try to parse `ephemeral` as a date if it looks like a date, which shouldn't be valid.

Also, `parse_date_param` returns a datetime, while `parse_number_of_days` returns an integer .. and `opt_ephemeral` expects a string, because it still calls `parse_number_of_days` later .

metrics_utility/management/validation.py
himdel: Still only validates one or the other. Shouldn't it be called for both?

metrics_utility/automation_controller_billing/helpers.py
```diff
@@ -9,7 +9,11 @@
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 
-from metrics_utility.exceptions import UnparsableParameter
+from metrics_utility.exceptions import DateFormatError, UnparsableParameter
+
+
+DATE_REGEX_PATTERN = r'^\d{4}-\d{2}(-\d{2})?$'
```
himdel: I wonder if this could be too restrictive?

Places using `dateutil.parser.parse` would previously also acccept dates like "4/5/2063", "15.5.2011", and such. 

Would it make sense to only use this for the month check? Then again I wonder if anyone is using these other formats, maybe it IS safe to require iso dates.

metrics_utility/management/commands/build_report.py
```diff
@@ -25,27 +37,39 @@ class Command(BaseCommand):
...
+        )
+        self.help_time_frame_extra_params = (
+            'Missing required parameter --month, --until, or --since. Metrics utility requires a value for at least \n'
+            'one of the following: month, since, until.'
+        )
+        self.help_month = 'Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report will \n'
+        "be generated if it doesn't exists already."
+        self.help_ephemeral = (
+            'Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report \n '
+            'will be generated if it does not exists already.'
```
himdel: this is the same as help_month .. different text for ephemeral?

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,98 @@
...
+        env_vars,
+        {'since': '2024-01-01', 'until': '2024-02-01', 'month': '2025-04', 'force': True},
+        'The --since and --until parameters are not allowed if the --month parameter is provided.',
+    )
+
+
+def test_renewal_guidancev2_fails_with_until_params():
+    command_args = [
+        {'args': {'until': '2025-01-01'}, 'report_type': 'RENEWAL_GUIDANCE'},
+        {'args': {'until': '2025-01-01'}, 'report_type': 'RENEWAL_GUIDANCEv2'},
```
himdel:
```suggestion
```

(guidance v2)

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,98 @@
...
+        {'args': {'until': '2025-01-01', 'since': '2024-02-01'}, 'report_type': 'RENEWAL_GUIDANCE'},
+    ]
+    for arg in command_args:
+        handle_build_exception(
+            {**env_vars, 'METRICS_UTILITY_REPORT_TYPE': arg['report_type']},
+            arg['args'],
+            'The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE or RENEWAL_GUIDANCEv2',
```
himdel:
```suggestion
            'The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE',
```

himdel: Also .. this test didn't fail, even when the code outputting the message was changed (https://github.com/ansible/metrics-utility/pull/128/files#diff-74815671e759671fde1bdd5f0276ffd27e215e903eb3dc74b3f3dc1da16c3380R95)

So .. might be worth investigating why it didn't fail when the message doesn't match now.

himdel (commented):
Getting closer :) 

I wonder if we could fix some of the remaining issues by having a definition table somewhere...

Such as

```
ccsp_build_report_params = {
  'since': {
    'required': false,
    'default': datetime.now(),
    'exclusive': ['month'],
    'format': [parse_datetime, parse_number_of_days]
  },
  'until': { ... }
}
```

to drive the various validator functions.

(Up to you though if you want to explore that or not.)

metrics_utility/management/commands/build_report.py
AlexSCorey: ok good point. I'll validate 'ephemeral' before we get down here.

metrics_utility/automation_controller_billing/helpers.py
AlexSCorey: No, I think you are right.  We should use parser.parse.  As for the month check, we have a current check happening [here](https://github.com/ansible/metrics-utility/pull/128/files/1a7be4180a36ccf06743818c0a46581ae3e6316b).  Do you think that will suffice?

AlexSCorey: So I ended up just removing that regex and opting for a new function to help validate expected date/time values.

cshiels-ie (approved):
Looks good to me

metrics_utility/management/validation.py
```diff
@@ -58,3 +62,61 @@ def handle_crc_ship_target(ship_target):
...
+            continue
+        try:
+            """Try to parse the date, and if it fails go to next loop iteration.  Then, determine if we need to render the failure message."""
+            re.match(DAY_OR_MONTH_REGEX_PATTERN, option)
+        except Exception:
+            exceptions.append('match')
+            help_text = getattr(self, f'help_{param}')
+            continue
+    if len(exceptions) and ('isdigit' or ('parser' and 'match') in exceptions):
+        raise UnparsableParameter(help_text)
```
himdel: * do not pass `self` to functions, only methods - but we don't need it here
* don't use `getattr` if you can prevent it, prefer hash lookups, or straight passing in the right value for the param
* `re.match` won't raise an exception, it just returns `None`
* `ephemeral` accepts "1d 2day 3days 4mo 5month 6months" only, `since` and `until` accept either "1d 2m 2033-03-03" (for gather) or "1d 2mo 3month 4months 5m 2066-06-06" (for build), `m` stands for minutes
* the function breaks the loop on some exception, but continues on others, but only ever returns UnparseableParameter with the help text of the last exception it handled ... why not just raise straight away

What about something like....

```suggestion
def handle_validate_date_param(value, help):
    if not value:
        return
        
    if value.isdigit():
        raise UnparsableParameter('Integers are not allowed for parameters --since and --until.')

    try:
        """Try to parse as date."""
        parser.parse(value)
        return
    except Exception:
        pass

    if re.match(DAY_OR_MONTH_REGEX_PATTERN, value):
        return

    raise UnparsableParameter(help)

def handle_validate_ephemeral_param(value, help):
    if not value:
        return
        
    if re.match(DAY_OR_MONTH_REGEX_PATTERN, value):
        return

    raise UnparsableParameter(help)
```

except that's just another way of spelling `parse_date_param` and `parse_number_of_days` which already exist? (assuming we added the isdigit there to the first one too)
(also `_parse_param` seems to do the same?)

metrics_utility/automation_controller_billing/helpers.py
```diff
@@ -9,7 +9,10 @@
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 
-from metrics_utility.exceptions import UnparsableParameter
+from metrics_utility.exceptions import DateFormatError, UnparsableParameter
+
+
+DAY_OR_MONTH_REGEX_PATTERN = r'.^\d+(d|day|days|m|mo|month|months)$'
```
himdel:
```suggestion
DAY_OR_MONTH_REGEX_PATTERN = r'^\d+(d|day|days|m|mo|month|months)$'
```

otherwise this requires a character before the start of the line

metrics_utility/management/commands/build_report.py
```diff
@@ -25,27 +36,37 @@ class Command(BaseCommand):
...
+            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+        )
+        self.help_time_frame_extra_params = (
+            'Missing required parameter --month, --until, or --since. Metrics utility requires a value for at least \n'
+            'one of the following: month, since, until.'
+        )
+        self.help_month = 'Month the report will be generated for, with format YYYY-MM. If this params is not provided, previous month report will \n'
+        "be generated if it doesn't exists already."
+        self.help_ephemeral = 'Duration in months or days to determine if host is ephemeral. Months are taken as 30days duration.\n'
+        'Example: --ephemeral=3months, or --ephemeral=3days'
```
himdel: This is fine as is if you're only using this as `self.help_something`, but if you want to use these with `getattr`, prefer something like

```
self.help = {
    'since': "Start date...",
    'until': "End date....",
    'month': "Mont the report....",
}
```

(and `self.help[option_name]`)

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +82,34 @@ def init_logging(self):
         handler.setLevel(logging.DEBUG)
         handler.setFormatter(logging.Formatter('%(message)s'))
         self.logger.addHandler(handler)
-        self.logger.propagate = False
-
-    def handle(self, *args, **options):
+        self.logger.propagate = True
```
himdel, line 88: is this change intentional? Looks unrelated? (was previously False)

metrics_utility/management/commands/build_report.py
```diff
@@ -25,27 +36,37 @@ class Command(BaseCommand):
...
-            "generated if it doesn't exists already.",
+    def __init__(self):
+        # help = 'Gather Automation Controller billing data'
+        self.help_since = (
+            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
         )
-        parser.add_argument('--since', dest='since', action='store', help='Date or number of days/months ago we want to generate the reports for.')
+        self.help_until = (
+            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+        )
```
himdel: 2m is never months, it's minutes.

2mo is months

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +82,34 @@ def init_logging(self):
...
+    def _parse_param(
+        self,
+        param_name,
+        options,
+    ):
+        param = options.get(param_name)
+        if (options.get('month') is not None or param is None) and param_name != 'ephemeral':
+            return None
+        if param:
+            return parse_date_param(param) if parser.parse(param) else parse_number_of_days(param)
```
himdel: * `parser.parse` raises an exception, or returns a truthy value, so putting it inside an `if` doesn't work.
* Let's keep `ephemeral` separate from `since` and `until`, the formats are different.
* the logic for ephmeral vs month is too complicated, and no longer used


maybe something like

```suggestion
    def _parse_date_param(self, value, month):
        if month is not None or value is None:
            return None

        return parse_date_param(value)
```

metrics_utility/management/validation.py
```diff
@@ -58,3 +62,61 @@ def handle_crc_ship_target(ship_target):
...
+    handle_validate_date_param(self, options)
+
+    until = parse_date_param(options.get('until'))
+    since = parse_date_param(options.get('since'))
+    report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE')
+    if report_type.startswith('RENEWAL_GUIDANCE') and until is not None:
+        raise BadParameter('The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE')
+
+    if since is None and until is None and opt_month is None:
+        raise MissingRequiredParameter(f"""{self.help_time_frame_extra_params} \n\n{self.help_since} \n{self.help_until} \n{self.help_month}""")
```
himdel: None of these are required. Month defaults to the last month when neither month nor since are provided.

..Except in renewal guidance report, where `--since` is required.

metrics_utility/management/commands/build_report.py
```diff
@@ -153,6 +165,24 @@ def handle(self, *args, **options):
...
+            MissingRequiredParameter,
+            UnparsableParameter,
+            BadParameter,
+            DateFormatError,
+        ) as e:
+            self.logger.error(str(e))
+            exit(1)
+        except Exception as e:
+            self.logger.exception(e)
+            exit(1)
```
himdel, line 186: `exit` from handle means `run_build_int` has to call `_handle` instead of `handle` now, same as `run_gather_int` already does

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,96 @@
...
+def test_invalid_month_format(caplog):
+    bad_inputs = ['abc', '12', 'mo3', '3mon', '3months', '3m', '3mo']
+
+    for bad_input in bad_inputs:
+        handle_build_exception(env_vars, {'month': bad_input}, 'Invalid --month format. Supported date format: YYYY-MM', caplog)
+
+
+def handle_build_exception(env_vars, params, error_message, caplog):
+    with caplog.at_level('ERROR'):
+        with pytest.raises(SystemExit) as exc_info:
```
himdel: `SystemExit` shouldn't happen once `run_build_int` is fixed to run `_handle` .. then you can test the real exception (but caplog is ok too)

(unless you're intentionally testing `handle()`?)

metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -67,6 +81,9 @@ def handle(self, *args, **options):
 
     def _handle(self, *args, **options):
         self.init_logging()
+
+        handle_validate_date_param(self, options)
```
himdel:
```suggestion
        handle_validate_date_param(options.get('since', None), self.help_since)
        handle_validate_date_param(options.get('until', None), self.help_until)
```

metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -27,7 +32,16 @@ class Command(BaseCommand):
...
     """
 
-    help = 'Gather Automation Controller billing data'
+    def __init__(self):
+        self.help_since = (
+            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of months (--since=2m).'
+        )
+        self.help_until = (
+            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of months (--until=2m).'
+        )
```
himdel:
```suggestion
        self.help_since = (
            'Start date for collection including (e.g. --since=2023-12-20), a number of days ago (--since=5d), or a number of minutes (--since=2m).'
        )
        self.help_until = (
            'End date for collection including (e.g. --until=2023-12-21), a number of days ago (--until=5d), or a number of minutes (--until=2m).'
        )
```

metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -27,7 +32,16 @@ class Command(BaseCommand):
...
+        self.help_time_frame_extra_params = (
+            'Missing required parameter --until, or --since. Metrics utility requires a value for at least one of the following: since, until.'
+        )
```
himdel: For gather, neither since nor until are required.

metrics_utility/management/validation.py
```diff
@@ -58,3 +62,61 @@ def handle_crc_ship_target(ship_target):
...
+            continue
+    if len(exceptions) and ('isdigit' or ('parser' and 'match') in exceptions):
+        raise UnparsableParameter(help_text)
+
+
+def validate_build_extra_params(self, options):
+    opt_month = options.get('month') or None
+    since = None
+    until = None
+    handle_validate_date_param(self, options)
```
himdel:
```suggestion
    handle_validate_date_param(options.get('since'), self.help_since)
    handle_validate_date_param(options.get('until'), self.help_until)
    handle_validate_ephemeral_param(options.get('ephemeral', None), self.help_ephemeral)
```

(except ephemeral is only used for renewal guidance)

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,96 @@
...
+import pytest
+
+from openpyxl import load_workbook
+from pandas import read_excel
+
+from metrics_utility import prepare
+from metrics_utility.test.util import run_build_int
+
+
+prepare()
```
himdel:
```suggestion
from metrics_utility.test.util import run_build_int


```

prepare is no longer needed in test_* files (post rebase)

metrics_utility/management/commands/gather_automation_controller_billing_data.py
himdel: Hm, except `handle_validate_date_param` calls `parse_date_param`, while gather otherwise uses `_handle_datelike` .. they probably should be merged together, but right now, they accept different formats

`build`:
```
    Nd (days)
    Nmo Nmonth Nmonths (months)
    Nm (minutes)
    YYYY-MM-DD
```

`gather`:
```
    Nd (days)
    Nm (minutes)
    YYYY-MM-DD
```

(merging the functions to support all those formats is an option, they're not incompatible .. I took a stab at it in https://github.com/ansible/metrics-utility/pull/136/commits/dd91d555751d6e1f6e70341009d81f22cb8ad737 if you'd like to go that way)

metrics_utility/management/commands/build_report.py
```diff
@@ -25,27 +36,37 @@ class Command(BaseCommand):
...
-    def add_arguments(self, parser):
-        parser.add_argument(
-            '--month',
-            dest='month',
-            action='store',
-            help='Month the report will be generated for, with format YYYY-MM. '
-            "If this params isn't provided, previou month report will be"
-            "generated if it doesn't exists already.",
+    def __init__(self):
+        # help = 'Gather Automation Controller billing data'
```
himdel:
```suggestion
```

stray comment from the other command?

himdel:
If it helps at all, here are my notes about the param values, types and formats:

```
build_report
  --month
    YYYY-MM or last month

    _handle_month returns "YYYY-MM", 2 date/datetimes without timezone (first of last month, first of this month)
    parsed as date-01 when present (datetime without timezone), defaults to previous month (date without timezone)
    ^=> opt_month (str), month (d/dt), next_month (d/dt)

    opt_month used for filenames / errors
    opt_month passed to ReportFactory as report_period
      ReportFactory overwrites it with extra_params report_period_range (from since/until) when present, passes to Report*
      Report* uses it to set a cell value
    month sets extra_params month_since (incl)
    month passed to extractor get_report_path - strftime to path (P/reports/Y/M)
    month passed to DataFrameEngineFactory as month
      passed to DataFrame*
      Base.dates
        uses month (with extra .replace(day=1)) to set beginning_of_the_month (or since_date)
        sets end_of_the_month from beg+months=1-days=1 (ignores month_until, off by a day from that) (or until_date)
        calls list_dates - each day between dates, including both
      self.dates called from all non-db dataframes
    next_month sets extra_params month_until (excl)
    extra_params passed to ExtractorFactory, Extractor* (before since_date/until_date/report_* get set, but after opt_since, opt_until, month_since, month_until)
      pass to extractors
      (uses only ship_path, opt_since (ExtractorControllerDB), bucket_* for s3, optional_sheets)
    extra_params passed to ReportSaverFactory
      pass to reportsavers
      (uses only report_spreadsheet_destination_path, bucket_* for s3)
    extra_params passed to DataframeEngineFactory
      use report_type, pass to dataframe
      Base.dates uses since_date, until_date, called from all non-db dataframes
      (uses only report_type, since_date, until_date)
    extra_params passed to ReportFactory
      use report_type, pass to Report
      ccspv2 _since_until uses month_since, month_until (and opt_since and opt_until)
        unsets timezone
        adds 1 day to opt_until only
        since/until wins over month, relies on unset until without since
      (renewal uses since_date, until_date, with pd.to_datetime)
      (uses optional_sheets, report_type, report_*, price_per_node, opt_since, month_since, opt_until, month_until, opt_ephemeral, since_date, until_date)
      ! opt_ephemeral only for renewal guidance

    ! --month is not supported for renewal guidance
      KeyError on report_period_range, since_date, until_date


  --since
    YYYY-MM-DD (or any parseable date format, doesn't overwrite hours or tz unless missing; (bare still allowed))
    Nd (start of day, utc)
    Nmo Nmonth Nmonths (start of day, utc)
    Nm (minutes, utc)

    parse_date_param returns a datetime with timezone (or None)
    ^=> opt_since (dtz | None)
    + extra_params since_date (d) - opt_since converted to date without timezone
      only when opt_since is set
    + extra_params report_period_range set to string of "since_date, until_date" (Y-M-D) ; only when opt_since is set
      also report_spreadsheet_destination_path

    extra_params opt_since passed to ExtractorFactory, Extractor*
      ExtractorControllerDB uses opt_since, passes to query,
      ! opt_since required for renewal guidance
    (extra_params passed to ReportSaverFactory)
    extra_params passed to DataframeEngineFactory
      Base.dates uses since_date, until_date, called from all non-db dataframes
    extra_params passed to ReportFactory
      ccspv2 _since_until uses opt_since & opt_until (and month_since, month_until); since/until wins over month, relies on unset until without since
      renewal uses since_date, until_date, with pd.to_datetime

  --until
    defaults to today
    ignored without --since
    not available for renewal guidance report

    YYYY-MM-DD (or any parseable date format, doesn't overwrite hours or tz unless missing; (bare still allowed))
    Nd (start of day, utc)
    Nmo Nmonth Nmonths (start of day, utc)
    Nm (minutes, utc)

    when missing and opt_since there, defaults to start of today utc

    parse_date_param returns a datetime with timezone (or None)
    ^=> opt_until (dtz | None) - only set when opt_since is
    + extra_params until_date (d) - opt_until converted to date without timezone
      but if opt_until was null (parse_date_param returned None), defaults to system time now, interpreted as utc, conveted to date without timezone
      only when opt_since is set
    + extra_params report_period_range set to string of "since_date, until_date" (Y-M-D) ; also report_spreadsheet_destination_path; only when opt_since is set

    (extra_params opt_until passed to ExtractorFactory, Extractor*)
    (extra_params passed to ReportSaverFactory)
    extra_params passed to DataframeEngineFactory
      Base.dates uses since_date, until_date, called from all non-db dataframes
    extra_params passed to ReportFactory
      ccspv2 _since_until uses opt_since & opt_until (and month_since, month_until); since/until wins over month, relies on unset until without since
      renewal uses since_date, until_date, with pd.to_datetime

  --ephemeral
    duration in months or days
    months treated as 30 days

    Nd Nday Ndays
    Nmo Nmonth Nmonths

    opt_ephemeral (str | None), passed to extra_params
    (extra_params passed to ExtractorFactory, Extractor*)
    (extra_params passed to ReportSaverFactory)
    (extra_params passed to DataframeEngineFactory)
    extra_params passed to ReportFactory
      ! opt_ephemeral only for renewal guidance

    ReportRenewalGuidance calls parse_number_of_days, returns int number of days (or throws), used for timedelta

  --force
    bool
    ends early when not set and file exists

  --verbose
    bool, unused, but needed for print_debug


gather_automation_controller_billing_data
  --dry-run
    bool
    ignored when --ship not present, error when both present

  --ship
    bool
    changes collection_type to MANUAL_COLLECTION (else DRY_RUN)
    affects shipping to c.r.c, cleanup of tmp tarballs, and updating last gather

  --since, --until
    both passed to _handle_interval, then _handle_datelike

    Nd (days)
    Nm (minutes)
    YYYY-MM-DD (or any parseable date format, doesn't overwrite hours or tz unless missing)

    disallows bare numbers
    otherwise quite similar to parse_date_param, just no long forms, or months
    returns a datetime with timezone (or None)
    ^=> since (dtz | None), until (dtz | None)

    passed to Collector.gather -> _gather_initialize -> _calculate_collection_interval
    until defaults to now when later than now
    since defaults to now when later than now

    if !since & !until
      set until to now (since unset)
    if !since & until
      nothing, since unset
    if since & !until
      set until to since+4weeks, or now, whichever is sooner
    if since & until
      if until is more than 4 weeks after since, set to since + 4weeks

    if since & since >= util, raise

    if since & since earlier than until-4weeks
      since = until-4weeks

    both since, until optional
    defaults
      4weeks ago to now when both empty
      until-4weeks when since missing
      since+4weeks when until missing (or now, when less than that .. but then since shifts to now-4weeks)

    ^=> gather_since (dtz | None), gather_until (dtz)
    then collection _gather_since uses the later value of until-4weeks and last_gather
```

Notably:

* `--month` is not supported for renewal guidance
* `--ephemeral` is only supported for renewal guidance
* `--since` is required only in renewal guidance

metrics_utility/management/commands/build_report.py
```diff
@@ -25,27 +36,37 @@ class Command(BaseCommand):
...
 
-    def add_arguments(self, parser):
-        parser.add_argument(
-            '--month',
-            dest='month',
-            action='store',
-            help='Month the report will be generated for, with format YYYY-MM. '
-            "If this params isn't provided, previou month report will be"
-            "generated if it doesn't exists already.",
+    def __init__(self):
```
himdel, line 43: When redefining the constructor, don't forget to call the constructor of the base class too...

metrics_utility/management/commands/gather_automation_controller_billing_data.py
AlexSCorey: Can we move this issue out of the scope of this pr?

metrics_utility/management/validation.py
AlexSCorey: Can we move this change `handle_validate_ephemeral_param(options.get('ephemeral', None), self.help_ephemeral)` to a separate issue?

metrics_utility/automation_controller_billing/helpers.py
```diff
@@ -9,7 +9,12 @@
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 
-from metrics_utility.exceptions import UnparsableParameter
+from metrics_utility.exceptions import DateFormatError, UnparsableParameter
+
+
+ALLOWED_EPHEMERAL_PATTERN = r'.^\d+(d|day|days|m|mo|month|months)$'
```
himdel:
```suggestion
ALLOWED_EPHEMERAL_PATTERN = r'^\d+(d|day|days|m|mo|month|months)$'
```

otherwise it never matches anything

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,132 @@
...
+
+    for bad_input in bad_inputs:
+        for arg in args:
+            cmd = Command()
+
+            help_text = cmd.help[arg]
+            if bad_input == '2':
+                help_text = 'Integers are not allowed for parameters --since and --until.'
+            e = handle_gather_exception(env_vars, {arg: bad_input}, UnparsableParameter)
+            e.name == help_text
```
himdel:
```suggestion
            assert e.name == help_text
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,132 @@
...
+    illegal_values = [
+        '1da',
+        '4mon',
+        '5mons',
+    ]
+    env_vars['METRICS_UTILITY_REPORT_TYPE'] = 'RENEWAL_GUIDANCE'
+
+    for value in illegal_values:
+        e = handle_build_exception(env_vars, {'ephemeral': value}, BadParameter)
+        e.name == 'METRICS_UTILITY_REPORT_TYPE RENEWAL_GUIDANCE does not allow --ephemeral.'
```
himdel:
```suggestion
        assert e.name == 'METRICS_UTILITY_REPORT_TYPE RENEWAL_GUIDANCE does not allow --ephemeral.'
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,132 @@
...
+
+    for bad_input in bad_inputs:
+        for arg in args:
+            cmd = Command()
+
+            help_text = cmd.help[arg]
+            if bad_input == '2':
+                help_text = 'Integers are not allowed for parameters --since and --until.'
+            e = handle_build_exception(env_vars, {arg: bad_input}, UnparsableParameter)
+            e.name == help_text
```
himdel:
```suggestion
            assert e.name == help_text
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,132 @@
...
+        e = handle_build_exception({**env_vars, 'METRICS_UTILITY_REPORT_TYPE': 'RENEWAL_GUIDANCE'}, arg['args'], BadParameter)
+        e.name = 'The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE'
+
+
+def test_invalid_month_format():
+    bad_inputs = ['abc', '12', 'mo3', '3mon', '3months', '3m', '3mo']
+
+    for bad_input in bad_inputs:
+        e = handle_build_exception(env_vars, {'month': bad_input}, DateFormatError)
+        e.name = 'Invalid --month format. Supported date format: YYYY-MM'
```
himdel:
```suggestion
        assert e.name == 'Invalid --month format. Supported date format: YYYY-MM'
```

metrics_utility/test/test_extra_params_validation.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def test_renewal_guidance_fails_with_until_params():
+    command_args = [
+        {'args': {'until': '2025-01-01'}},
+        {'args': {'until': '2025-01-01', 'since': '2024-02-01'}},
+    ]
+    for arg in command_args:
+        e = handle_build_exception({**env_vars, 'METRICS_UTILITY_REPORT_TYPE': 'RENEWAL_GUIDANCE'}, arg['args'], BadParameter)
+        e.name = 'The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE'
```
himdel:
```suggestion
        assert e.name == 'The --until parameter is not allowed when environment variable METRICS_UTILITY_REPORT_TYPE is RENEWAL_GUIDANCE'
```

himdel (commented):
a couple of missing asserts / assignments instead of  comparisons in tests

metrics_utility/management/commands/build_report.py
```diff
@@ -61,43 +87,34 @@ def init_logging(self):
...
+    def _parse_param(
+        self,
+        param_name,
+        options,
+    ):
+        param = options.get(param_name)
+        if (options.get('month') is not None or param is None) and param_name != 'ephemeral':
+            return None
+        if param:
+            return parse_date_param(param) if parser.parse(param) else parse_number_of_days(param)
```
himdel:
```suggestion
    def _parse_param(
        self,
        param_name,
        options,
    ):
        param = options.get(param_name)
        if (options.get('month') is not None or param is None):
            return None
        return parse_date_param(param)
```

* because `param_name != 'ephemeral'` is always true, we never call it with that param_name now
* if param is None, we return none in the first if already
* and `parse_number_of_days` returns and integer number of days, where since & until expect a datetime, which is handled by `parse_data_param` (`parse_number_of_days` is just for ephemeral)

metrics_utility/management/validation.py
```diff
@@ -110,3 +119,97 @@ def handle_not_crc():
...
+
+    if since:
+        raise MissingRequiredParameter(f"""{help_text.time_frame_extra_params} \n\n{since_help} \n{until_help} \n{help_text.month}""")
+
+
+def handle_validate_ephemeral_param(value, help):
+    report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE', None)
+    if not value:
+        return
+    if report_type != report_type.startswith('RENEWAL_GUIDANCE'):
```
himdel:
```suggestion
    if not report_type.startswith('RENEWAL_GUIDANCE'):
```

(report_type != (True|False)) is always true)

OR

```suggestion
    if report_type != 'RENEWAL_GUIDANCE':
```

PR #135, opened 2025-06-06
AAP-45503- Better Error Handling for enviroment variables 

cshiels-ie:
[[AAP-45503]](https://issues.redhat.com/browse/AAP-45503)

## Description

Better Error Handling for enviroment variables

- What is being changed?
Added in validate location for verifying all variables so this can be done cleanly in one place

- Why is this change needed?
At the moment environment variable are verified or not even verified once they are run causing exceptions to be thrown without the full details. this way the environment variables will be checked at first run and show a error and helper. This allows multiple issues to be caught at first run instead of individually. 

Resolves the issue and gives more structure going forward. 

##  Changes 

- Added Try/Except ti the main manage.py to gracefully exit in event of a MissingRequiredEnvVar Exception 
[manage.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-57f701b24994e3ca4194f54015c59a3bc690a0f9a9c23617980b952c979fe8b7)
- Added handle_env_validation to Build Report [metrics_utility/management/commands/build_report.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-216a16965cc1fe989bdf35aef44e5ecd26f303ce3dd23812ff00862c8bc662a3)
- Addition to Validation.py of the handle_env_validation method and constants of valid values [metrics_utility/management/validation.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-74815671e759671fde1bdd5f0276ffd27e215e903eb3dc74b3f3dc1da16c3380)
- Added basic unit tests for code coverage for [metrics_utility/test/ccspv_reports/test_build_report_unit.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-2df7e5d5f20f0d957ef9210f8a4e96a8e045444087bd770b2e3217e1ec00b3ab)
- Added unit tests for [test_validation.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-18504e9f84b6305d6410e1e8b1cc3f822e8ea643a0b8160566c0ce40f85a430c)


metrics_utility/management/validation.py
```diff
@@ -4,6 +4,37 @@
...
+        'inventory_scope',
+        'usage_by_organizations',
+        'usage_by_collections',
+        'usage_by_roles',
+        'usage_by_modules',
+        'managed_nodes_by_organization',
+        'data_collection_status',
+    },
+}
+VALID_COLLECTORS = {'main_host', 'main_jobevent', 'main_indirectmanagednodeaudit'}
```
AlexSCorey, line 34: `main_hostmetric` is a valid collector too.

metrics_utility/management/validation.py
```diff
@@ -98,6 +129,169 @@ def handle_crc_ship_target():
...
+
+    Args:
+        errors (list): A list to which error messages will be appended if validation fails.
+
+    Returns:
+        str or None: The value of the 'METRICS_UTILITY_REPORT_TYPE' environment variable if set, otherwise None.
+    """
+    report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE', None)
+    if report_type and report_type not in VALID_REPORT_TYPES:
+        errors.append(
```
AlexSCorey, line 147: TIL what a soft validation pattern is.

metrics_utility/management/validation.py
```diff
@@ -98,6 +129,169 @@ def handle_crc_ship_target():
...
+        ship_target (str): The value of the METRICS_UTILITY_SHIP_TARGET environment variable.
+
+    Notes:
+        - For 'directory' ship target, checks if METRICS_UTILITY_SHIP_PATH is an existing directory.
+        - Appends an error message to 'errors' if the directory does not exist.
+    """
+    ship_path = os.getenv('METRICS_UTILITY_SHIP_PATH', None)
+    if ship_target == 'directory' and ship_target:
+        if not os.path.isdir(ship_path):
+            errors.append(f'Invalid METRICS_UTILITY_SHIP_PATH: {ship_path} is not an existing directory.')
```
AlexSCorey, line 257: I'm pretty sure we have functionality somewhere to create the dir if it doesn't exist. Of course that requires that the path provided to the new dir is a valid path, and doesn't link to a file somewhere in the middle of the path.

IE, I'm not sure we want this to fail at the very beginning because somewhere in our execution we create a dir if needed.

AlexSCorey (approved):
Looks real good.  Just a few comments/things to consider.

metrics_utility/management/validation.py
cshiels-ie: I think main_hostmetric relates to the table. We pull the data from but isn't a collector option as far as I can tell

metrics_utility/management/validation.py
cshiels-ie: This isn't soft validation tho the errors list gets loaded into an exception that will stop the service

metrics_utility/management/validation.py
cshiels-ie: Thanks I'll double check this :)

PR #137, opened 2025-06-09
 AAP-45503- Better Error Handling for enviroment variables

cshiels-ie:
[[AAP-45503]](https://issues.redhat.com/browse/AAP-45503)

## Description

Better Error Handling for enviroment variables

- What is being changed?
Added in validate location for verifying all variables so this can be done cleanly in one place

- Why is this change needed?
At the moment environment variable are verified or not even verified once they are run causing exceptions to be thrown without the full details. this way the environment variables will be checked at first run and show a error and helper. This allows multiple issues to be caught at first run instead of individually. 

Resolves the issue and gives more structure going forward. 

##  Changes 

- Added handle_env_validation to Build Report [metrics_utility/management/commands/build_report.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-216a16965cc1fe989bdf35aef44e5ecd26f303ce3dd23812ff00862c8bc662a3)
- Added handle_env_validation to Gather Method [metrics_utility/management/commands/gather_automation_controller_billing_data.py](https://github.com/ansible/metrics-utility/pull/137/files#diff-bb98f7ff0b0f4e4c66ebfeb4296c00cf73d8e86eafcb1fd12ba5c7eca4aeb2a7)files#diff-216a16965cc1fe989bdf35aef44e5ecd26f303ce3dd23812ff00862c8bc662a3)
- Addition to Validation.py of the handle_env_validation method and constants of valid values [metrics_utility/management/validation.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-74815671e759671fde1bdd5f0276ffd27e215e903eb3dc74b3f3dc1da16c3380)
- Added basic unit tests for code coverage for [metrics_utility/test/ccspv_reports/test_build_report_unit.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-2df7e5d5f20f0d957ef9210f8a4e96a8e045444087bd770b2e3217e1ec00b3ab)
- Added unit tests for [test_validation.py](https://github.com/ansible/metrics-utility/pull/135/files#diff-18504e9f84b6305d6410e1e8b1cc3f822e8ea643a0b8160566c0ce40f85a430c)

Closed previous PR due to unsigned commits https://github.com/ansible/metrics-utility/pull/135

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,7 +1,16 @@
 #!/usr/bin/env python
 """Django's command-line utility for administrative tasks."""
 
+import sys
+
+from metrics_utility.exceptions import MissingRequiredEnvVar
+
+
 if __name__ == '__main__':
     from metrics_utility import manage
 
-    manage()
+    try:
+        manage()
+    except MissingRequiredEnvVar as exc:
+        print(f'\nEnvironment Variable Validation Error: \n{exc}\n')
+        sys.exit(1)
--- a/manage.py
+++ b/manage.py
@@ -1,16 +1,7 @@
 #!/usr/bin/env python
 """Django's command-line utility for administrative tasks."""
 
-import sys
-
-from metrics_utility.exceptions import MissingRequiredEnvVar
-
-
 if __name__ == '__main__':
     from metrics_utility import manage
 
-    try:
-        manage()
-    except MissingRequiredEnvVar as exc:
-        print(f'\nEnvironment Variable Validation Error: \n{exc}\n')
-        sys.exit(1)
+    manage()
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -32,7 +32,8 @@ VALID_SHEETS = {
     },
 }
 VALID_COLLECTORS = {'main_host', 'main_jobevent', 'main_indirectmanagednodeaudit'}
-VALID_SHIP_TARGET = {'directory', 's3', 'crc'}
+VALID_SHIP_TARGET_BUILD = {'directory', 's3'}
+VALID_SHIP_TARGET_GATHER = {'directory', 's3', 'crc'}
 
 
 logger = logging.getLogger(__name__)
@@ -216,7 +217,7 @@ def validate_collectors(errors):
             errors.append(f'Invalid METRICS_UTILITY_OPTIONAL_COLLECTORS: {", ".join(invalid)}. Valid values: {", ".join(VALID_COLLECTORS)}')
 
 
-def validate_ship_target(errors):
+def validate_ship_target(errors, ship_target_type):
     """
     Validates the 'METRICS_UTILITY_SHIP_TARGET' environment variable against a set of valid ship targets.
 
@@ -234,8 +235,8 @@ def validate_ship_target(errors):
         - Error messages include the invalid ship target and the list of valid values.
     """
     ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET', None)
-    if ship_target and ship_target not in VALID_SHIP_TARGET:
-        errors.append(f'Invalid METRICS_UTILITY_SHIP_TARGET: {ship_target}. Valid values: {", ".join(VALID_SHIP_TARGET)}')
+    if ship_target and ship_target not in ship_target_type:
+        errors.append(f'Invalid METRICS_UTILITY_SHIP_TARGET: {ship_target}. Valid values: {", ".join(ship_target_type)}')
     return ship_target
 
 
@@ -257,7 +258,7 @@ def validate_ship_path(errors, ship_target):
             errors.append(f'Invalid METRICS_UTILITY_SHIP_PATH: {ship_path} is not an existing directory.')
 
 
-def handle_env_validation():
+def handle_env_validation(method: str):
     """
     Validates required environment variables and configuration for the application.
 
@@ -286,7 +287,10 @@ def handle_env_validation():
     report_type = validate_report_type(errors)
     validate_ccsp_report_sheets(errors, report_type)
     validate_collectors(errors)
-    ship_target = validate_ship_target(errors)
+    if method == 'gather':
+        ship_target = validate_ship_target(errors, VALID_SHIP_TARGET_GATHER)
+    else:
+        ship_target = validate_ship_target(errors, VALID_SHIP_TARGET_BUILD)
     validate_ship_path(errors, ship_target)
     if errors:
         raise MissingRequiredEnvVar('\n'.join(errors))
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -4,6 +4,37 @@ import os
 from metrics_utility.exceptions import MissingRequiredEnvVar
 
 
+# Constants for valid values
+VALID_REPORT_TYPES = {'CCSP', 'CCSPv2', 'RENEWAL_GUIDANCE'}
+VALID_SHEETS = {
+    'CCSP': {
+        'ccsp_summary',
+        'managed_nodes',
+        'indirectly_managed_nodes',
+        'inventory_scope',
+        'usage_by_collections',
+        'usage_by_roles',
+        'usage_by_modules',
+        'usage_by_organizations',
+    },
+    'CCSPv2': {
+        'ccsp_summary',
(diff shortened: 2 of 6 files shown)


manage.py
```diff
@@ -1,7 +1,16 @@
...
+
 if __name__ == '__main__':
     from metrics_utility import manage
 
-    manage()
+    try:
+        manage()
+    except MissingRequiredEnvVar as exc:
+        print(f'\nEnvironment Variable Validation Error: \n{exc}\n')
+        sys.exit(1)
```
himdel: This bit probably needs merging with the other places doing the same thing...

`gather_*` already does this in the command - https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/management/commands/gather_automation_controller_billing_data.py#L63-L72

`build_report` doesn't yet, but does after Alex' PR - https://github.com/ansible/metrics-utility/pull/128/files#diff-216a16965cc1fe989bdf35aef44e5ecd26f303ce3dd23812ff00862c8bc662a3R167-R183

I do agree it makes sense to do it here, rather than in the individual commands, so we should probably move all those handle wrappers here, remove them from individual commands, and remove the workaround for gather tests - https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/test/util.py#L70-L71 - and call handle again there.

(But if you'd prefer to do that in a separate PR after all these are merged, feel free, this works as is :))

cshiels-ie: > This bit probably needs merging with the other places doing the same thing...
> 
> `gather_*` already does this in the command - https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/management/commands/gather_automation_controller_billing_data.py#L63-L72
> 
> `build_report` doesn't yet, but does after Alex' PR - https://github.com/ansible/metrics-utility/pull/128/files#diff-216a16965cc1fe989bdf35aef44e5ecd26f303ce3dd23812ff00862c8bc662a3R167-R183
> 
> I do agree it makes sense to do it here, rather than in the individual commands, so we should probably move all those handle wrappers here, remove them from individual commands, and remove the workaround for gather tests - https://github.com/ansible/metrics-utility/blob/devel/metrics_utility/test/util.py#L70-L71 - and call handle again there.
> 
> (But if you'd prefer to do that in a separate PR after all these are merged, feel free, this works as is :))

Yeah good point! I think I should remove it from this PR and we can do it as part of a refactor and instead of just MissingRequiredEnvVar we do it for all the different exceptions.

metrics_utility/management/validation.py
```diff
@@ -4,6 +4,37 @@
...
+        'usage_by_organizations',
+        'usage_by_collections',
+        'usage_by_roles',
+        'usage_by_modules',
+        'managed_nodes_by_organization',
+        'data_collection_status',
+    },
+}
+VALID_COLLECTORS = {'main_host', 'main_jobevent', 'main_indirectmanagednodeaudit'}
+VALID_SHIP_TARGET = {'directory', 's3', 'crc'}
```
himdel:
```suggestion
VALID_SHIP_TARGET = {'directory', 's3', 'crc', 'controller_db'}
```

(though `crc` is only available in gather, not build_report)

himdel:
What happens to `gather_automation_controller_billing_data` - this PR only deals with `build_report` now, I think the issue is for both?

(Perfectly reasonable to split into multiple PRs, just checking what your plan is with the second command :))

cshiels-ie:
> `gather_automation_controller_billing_data`



> What happens to `gather_automation_controller_billing_data` - this PR only deals with `build_report` now, I think the issue is for both?
> 
> (Perfectly reasonable to split into multiple PRs, just checking what your plan is with the second command :))



> What happens to `gather_automation_controller_billing_data` - this PR only deals with `build_report` now, I think the issue is for both?
> 
> (Perfectly reasonable to split into multiple PRs, just checking what your plan is with the second command :))

I think I missed that it should be effecting `gather_automation_controller_billing_data`. I think I'll pull it into this PR it shouldn't be a lot extra.

himdel:
Sounds good to me :) Thanks

metrics_utility/management/validation.py
```diff
@@ -4,6 +4,38 @@
...
+        'usage_by_collections',
+        'usage_by_roles',
+        'usage_by_modules',
+        'managed_nodes_by_organization',
+        'data_collection_status',
+    },
+}
+VALID_COLLECTORS = {'main_host', 'main_jobevent', 'main_indirectmanagednodeaudit'}
+VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
+VALID_SHIP_TARGET_GATHER = {'directory', 's3', 'crc', 'controller_db'}
```
himdel: In that case, it's

```suggestion
VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
VALID_SHIP_TARGET_GATHER = {'directory', 's3', 'crc'}
```

metrics_utility/management/validation.py
```diff
@@ -98,6 +130,177 @@ def handle_crc_ship_target():
...
+    Args:
+        errors (list): A list to which error messages will be appended if validation fails.
+        ship_target (str): The value of the METRICS_UTILITY_SHIP_TARGET environment variable.
+
+    Notes:
+        - For 'directory' ship target, checks if METRICS_UTILITY_SHIP_PATH is an existing directory.
+        - Appends an error message to 'errors' if the directory does not exist.
+    """
+    ship_path = os.getenv('METRICS_UTILITY_SHIP_PATH', None)
+    if ship_target == 'directory' and ship_target:
```
himdel:
```suggestion
    if ship_target in ['directory', 'controller_db']:
```

(`s3` doesn't use a local ship path, `crc` doesn't use ship path at all, but `controller_db` does - a local dir for output)

himdel: Also .. the "path exists" check might only make sense for `build_report`.

For gather, we call `os.makedirs` in `PackageDirectory` to create it.

metrics_utility/management/validation.py
```diff
@@ -98,6 +130,177 @@ def handle_crc_ship_target():
...
+    an error message is appended to the provided errors list.
+
+    Args:
+        errors (list): A list to which error messages will be appended if validation fails.
+
+    Returns:
+        str or None: The value of the 'METRICS_UTILITY_REPORT_TYPE' environment variable if set, otherwise None.
+    """
+    report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE', None)
+    if report_type and report_type not in VALID_REPORT_TYPES:
```
himdel, line 156: are we intentionally ignoring the "no value" scenario here?

If the value is None, `handle_env_validation` won't complain about it, but maybe it should?

EDIT: `METRICS_UTILITY_REPORT_TYPE` is required for `build_report`, but not used at all for gather

metrics_utility/management/validation.py
cshiels-ie: Perfect I fixed them up now 👍

metrics_utility/management/validation.py
cshiels-ie: I have fixed that now do it allows none for Gather method

metrics_utility/management/validation.py
```diff
@@ -107,6 +139,190 @@ def handle_crc_ship_target():
...
+        errors.append(
+            f'Invalid METRICS_UTILITY_REPORT_TYPE: {report_type}. Valid values: {", ".join(VALID_REPORT_TYPES)}. '
+            f'Please note these values are case sensitive'
+        )
+        if method == 'build' and report_type is None:
+            errors.append(
+                f'Invalid METRICS_UTILITY_REPORT_TYPE is Empty. Valid values: {", ".join(VALID_REPORT_TYPES)}. '
+                f'Please note these values are case sensitive'
+            )
```
himdel:
```suggestion
    if method == 'build' and report_type is None:
        errors.append(
            f'Invalid METRICS_UTILITY_REPORT_TYPE is Empty. Valid values: {", ".join(VALID_REPORT_TYPES)}. '
            f'Please note these values are case sensitive'
        )
```

(inside the top if, report_type is never None)

himdel (approved):
LGTM :+1: :)

PR #143, opened 2025-06-17
AAP-47614 - Milan tests jobs hosts gathering

MilanPospisil:
[AAP-47614]

## Description

Goal is to generate data directly using SQL and then write test for gather billing data and verify that returned csv of job host summary is correct.

IMPORTANT!!!
Test is disabled for now, first PR has to be merged, because pytest.yml has pull request target, thus, any changes will become real after PR is merged.

- pytest.yml is enabling the SQL insert in CI (it is not working until it is merged)

- docker-compose.yaml is calling the script for data initialization

- the script can be also called using make (change in Makefile) for testing purposes - but you need to enable random suffix again so the repeatedly inserted data are not in naming conflict - see in sql script variables declaration. Otherwise random suffix is set to constant variable so our db init always produce the same data for stable predictable tests

- main_jobhostsummary.sql is long script for inserting all the data we need for gathering. It is mocked data that are suitable only for the gathering tests (and maybe performance tests), it is not aiming to be representant of the real controller db. Better data can be generated by running playbooks, but that is quite slow, this directly SQL Insert is much faster and better suited for the tests.

- test_hobhostsummary_gather.py is the test that is disabled for now, it will be enabled in future PR. You can test it by removing return and comments covering the code

## Testing

You must use standalone metrics utility version, not the one running inside awx container.

Ensure your db is up and running:
docker compose -f tools/docker/docker-compose.yaml up

Run the pytest itself for this file only:
uv run pytest -s metrics_utility/test/gather/test_jobhostsummary_gather.py

### Steps to Test

1.
2.
3.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers


ShaiahWren (commented):
Hey Milan!
Nice, it looks like hosts_variables query runs once per host and should satisfy the DISTINCT keyword

Q: Is this meant to be a correctness test or performance test? test_jobhostsummary_gather.py
If it is meant to test the difference between old and new query, explicit timing and assertions would be ideal. 

Could this test check for edge cases such as no data present in the query, extra-long strings, or a host with no var?

MilanPospisil:
> Hey Milan! Nice, it looks like hosts_variables query runs once per host and should satisfy the DISTINCT keyword
> 
> Q: Is this meant to be a correctness test or performance test? test_jobhostsummary_gather.py If it is meant to test the difference between old and new query, explicit timing and assertions would be ideal.
> 
> Could this test check for edge cases such as no data present in the query, extra-long strings, or a host with no var?

There is no optimization of SQL query yet, for this I need the test to ensure I did not broke anything. So this issue will have 3 PRs:

1) (This PR) - this will merge the pytest.yaml that inserts testing data into database - it must be merged first in order to pytest.yaml changes work, because it is pull request target workflow = it will start to work after it is merged

2) Test PR - after pytest.yaml merged and generating data into the database, test will be eneabled and assert the expected and actual gathered values are the same

3) Optimization PR - This will merge the optimizations itself and using the test in previous PR, it will assert new values are still ok.

No time assertions will be there (at least for now), this is part of performance tests that are not part of any CI tests, because they are slow.

metrics_utility/test/gather/test_jobhostsummary_gather.py
```diff
@@ -0,0 +1,105 @@
...
+]
+
+
+@pytest.mark.filterwarnings('ignore::ResourceWarning')
+def test_command(cleanup_glob):
+    """Build xlsx report using build command and test its contents."""
+    # Enable it by the second PR, first this PR has to be merged in order to
+    # pytest.yml works, because it is pull request target
+    return
+    """
```
himdel: can you just `@pytest.skip` instead?

Or move the test to a separate PR?

There's not much point adding a test that's disabled, what about just adding the CI stuff here?

himdel (approved):

PR #141, opened 2025-06-12
AAP-47719 - Fixing --Help command Bug

cshiels-ie:
Fixing --help command broken by recent PR. Added tests to avoid this error in the future and clean up some code. 

Link to Issue https://issues.redhat.com/browse/AAP-47719

PR #146, opened 2025-06-18
Commands: move handle_env* *after* init_logging

himdel:
Follows #141, issue AAP-47719
Issue: AAP-47951

Moving `handle_env_validation` inside `handle()` fixed `--help` breaking,
but introduced

```
   File "/home/himdel/metrics/metrics-utility/metrics_utility/management/commands/build_report.py", line 206, in handle
    self.logger.error(str(e))
    ^^^^^^^^^^^
AttributeError: 'Command' object has no attribute 'logger'
```

when running without `--help` (and when not all env variables were valid/provided)

(And updating the S3 env validation test to cope with `handle_env_validation` being a part of `handle()` now.)

### DIFF
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -102,6 +102,7 @@ class Command(BaseCommand):
     def _handle(self, *args, **options):
         self.init_logging()
 
+        handle_env_validation('build')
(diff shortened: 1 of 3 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #142, opened 2025-06-16
Param validation fixups

himdel:
Issue: AAP-47744
Follows #128 

When running `build_report` with `METRICS_UTILITY_REPORT_TYPE=RENEWAL_GUIDANCE`:

* allow `--since`
* disallow `--month` 
* disallow `--until`

When running `build_report` with `METRICS_UTILITY_REPORT_TYPE=CCSP` or `CCSPv2`:
* disallow `--until` *without* `--since`

PR #147, opened 2025-06-18
build_report: default month to last month

himdel:
Follows #128, #142 
Issue: AAP-47953

Set `opt_month` even when `options['month']` was empty - the default for `--month` is the last month.
(The value gets ignored when `--since` is used, but used otherwise.)

Fixes:

```
$ METRICS_UTILITY_REPORT_TYPE=CCSPv2 METRICS_UTILITY_SHIP_TARGET=directory METRICS_UTILITY_SHIP_PATH="./metrics_utility/test/test_data/" uv run ./manage.py build_report 
Automation Controller modules not found in /awx_devel (AWX_PATH). Using mock and continuing.
No billing data for month None
```

(expected `No billing data for month 2025-05`)

(Also affects the filename of the report if there are data, and the reported report period.)

### DIFF
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -111,10 +111,10 @@ class Command(BaseCommand):
 
     def _handle(self, *args, **options):
         self.init_logging()
-        og_month, month, next_month = handle_month(options.get('month') or None)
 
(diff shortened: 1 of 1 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #136, opened 2025-06-09
Cleanups around params

himdel:
This is a collection of semi-related commits, some work in progress, some ready, some suggestions, some fixes, some ideas, vaguely related to AAP-43196 and tasks under it - environmental variable and command line argument validation.

Currently contains a mix of:
* follow-ups to the epic, around making `--help` responsive to report type, moving exception handling to a base class
* dead/duplicate code cleanups / refactorings in related code, simplify logic around extra_params values
* obsolete help suggestions (which *might* still have bits that we'd want to add)
* extending the formats accepted for relative dates and intervals
* fixing up more edge cases with param checking
* test cleanups where we're skipping required params or providing extraneous params

---

The PR is never meant to be merged as a whole, I'll be pulling out bits and pieces that make sense together into individual PRs over time.

PR #149, opened 2025-06-18
`report_period` - move inside `extra_params`, merge with `report_period_range`

himdel:
`report_period` was always set to `opt_month`,
but `ReportFactory` would overwrite it with `report_period_range` when present (when `--since`),
and renewal guidance would ignore it and only use `report_period_range` (and `KeyError` for `--month`)

=> always use `report_period` now, set it to `opt_month` for `--month`, and to original `report_period_range` value in the `--since` branch

And move `report_period` inside `extra_params`, like `report_period_range` was:

`ReportFactory`: `remove report_period` & `ship_target` params
  `ship_target` is unused
  `report_period` moves inside `extra_params`

`Report*`: remove `report_period` constructor param, load from `extra_params`

`build_report`: set `extra_params['report_period']`, don't pass `report_period` & `ship_target`

tests - remove `report_period` param, rename `report_period_range` extra param to `report_period`

And fix a `ship_target` typo in dataframe engine factory - should be `report_type`

(pulling out cleanups from #136; this should behave exactly the same as before)

### DIFF
--- a/metrics_utility/automation_controller_billing/dataframe_engine/factory.py
+++ b/metrics_utility/automation_controller_billing/dataframe_engine/factory.py
@@ -12,26 +12,26 @@ class Factory:
         self.month = month
         self.extra_params = extra_params
 
-        self.report_type = extra_params['report_type']
-
     def create(self):
-        if self.report_type == 'CCSP':
+        report_type = self.extra_params['report_type']
+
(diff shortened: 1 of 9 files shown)


MilanPospisil (approved):
LGTM :+1: 

Well done, we are moving forward to clear this parameter mess :-)

PR #150, opened 2025-06-18
Unify exception handling between commands

himdel:
Follows #137 

Change metrics utility exceptions to inherit from `MetricsException`,
move the try/catch from both commands' `handle()` method to the caller - `ManagementUtility`.

Thus:
* `command.handle` returns an unhandled exception, that's either a bug, or inherits from `MetricsException`
* `management_utility` handles `MetricsException` by printing out a nice message, and `Exception` by printing out the whole stack trace; exit(1) in both cases
* tests no longer mock `_handle`, and don't see a `SystemExit` except for `--help`, getting a `MetricsException` instead

### DIFF
--- a/metrics_utility/exceptions.py
+++ b/metrics_utility/exceptions.py
@@ -1,48 +1,43 @@
-class BadShipTarget(Exception):
+class MetricsException(Exception):
     def __init__(self, message):
         self.name = message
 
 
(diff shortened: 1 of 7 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #152, opened 2025-06-19
Replace pytz with datetime.timezone

himdel:
Code scanning complains about using `pytz` with python >=3.9 (https://github.com/ansible/metrics-utility/security/code-scanning/1 & https://github.com/ansible/metrics-utility/security/code-scanning/2)

We're using only `pytz.utc`, and passing it to `datetime.datetime`
=> replace with `datetime.timezone.utc`

(and remove the pytz dependency)


AlexSCorey (approved):
Looks good.

PR #153, opened 2025-06-24
Cleanups around `extra_params`

himdel:
More cleanups around `extra_params`, follows #149, pulls out more fixups from #136

`get_report_path`: only used in `build_report`, move there, replaces 3 copies of the same with 1
`get_path_prefix`: move to extractor base class (and pass extra_params), replaces 2 copies with 1, removes dead artifact in controller db
`ReportSaver`s: cosmetic, replaces `True if condition else False` with `condition`
`get_optional_collectors`: only used in `collectors`, move there, removes dead artifact in build report


AlexSCorey (approved):
I like these changes.  Thanks for putting this together.

PR #154, opened 2025-06-24
Test fixups

himdel:
Pulling out changes from #136 - test fixups related to param validation

* don't pass `ephemeral` in ccsp tests, not applicable (where not testing for failure when provided)
* don't pass `verbose: False`, is the default
* pass valid values in `ephemeral` - no spaces allowed (yet)
* add report type to build validator tests (where not testing for failure when missing), change shared `month` to `since` to avoid renewal breaking


AlexSCorey (approved):
Looks good to me.

PR #155, opened 2025-06-27
`build_report`: pass parsed `ephemeral_days` in `extra_params`

himdel:
Follows #149, #153 - continuing the theme of cleaning up `extra_params`

Moving the `--ephemeral` parsing step closer to param validation, and renaming the variable to be clearer about what the value is.

before:
  `opt_ephemeral` gets passed as is to `ReportRenewalGuidance`
  `ReportRenewalGuidance` calls `parse_number_of_days` on it

after:
  `build_report` calls `parse_number_of_days`
  `ReportRenewalGuidance` receives parsed `ephemeral_days` (number or None)

(and left-shift an else block after true block returns in renewal guidance)

`parse_number_of_days`: allow `None`, return `None` - symmetrical with what `parse_date_param` does
`parse_date_param`: simplify with an early return

And add tests for the parse functions.


metrics_utility/automation_controller_billing/report/report_renewal_guidance.py
```diff
@@ -111,26 +111,24 @@ def build_spreadsheet(self):
     def df_managed_nodes_query(self, dataframe, ephemeral=None, with_deleted=False):
         if ephemeral is None:
             return dataframe[~dataframe['deleted']]
-        else:
-            # Take only non deleted
-            if not with_deleted:
-                dataframe = dataframe[~dataframe['deleted']]
-
-            # Filter ephemeral based on number of automated days
-            ephemeral_days = parse_number_of_days(self.extra_params.get('opt_ephemeral'))
```
MilanPospisil, line 120: Where this dissapeared? I don't see this in newly added code below.

Edit: I found it in another file!

metrics_utility/automation_controller_billing/report/report_renewal_guidance.py
```diff
@@ -111,26 +111,24 @@ def build_spreadsheet(self):
...
-            if ephemeral is False:
-                return dataframe[(dataframe['days_automated'] > ephemeral_days) | (dataframe['first_automation'] > ephemeral_threshold)]
+        # Take only non deleted
+        if not with_deleted:
+            dataframe = dataframe[~dataframe['deleted']]
+
+        # Ephemeral threshold, host's first automation must be older than ephemeral threshold
+        # to be considered as ephemeral
+        ephemeral_threshold = (
+            pd.to_datetime(datetime.datetime.now() - datetime.timedelta(days=self.ephemeral_days - 1), format='ISO8601')
```
MilanPospisil, line 122: Maybe you intended to use now() from helpers everywhere including here?

metrics_utility/management/commands/build_report.py
```diff
@@ -106,13 +107,14 @@ def handle(self, *args, **options):
         opt_month, month, next_month = handle_month(options.get('month') or None)
         opt_since = parse_date_param(options.get('since'))
         opt_until = parse_date_param(options.get('until'))
+        opt_ephemeral = parse_number_of_days(options.get('ephemeral'))
```
MilanPospisil, line 110: I hear Sheldon saying: Why? Why? Oh, thats why! :-)

MilanPospisil:
Great Job!

Once the datetime.now() is replaced with helpers.now(), you can merge it!

MilanPospisil (approved):
LGTM :+1: 

Dont forget to replace datetime.now with helpers.now

Ladas (approved):
👍

metrics_utility/automation_controller_billing/report/report_renewal_guidance.py
MilanPospisil: And can you please add some tests for the ephemeral where the datetime.now() is being tested by mocking? This should be catch by tests.

himdel: No. That `now()` function is for mocking the date for the helpers, there's no reason to mock this one given the tests we have.

I'm not trying to write tests for renewal guidance here, just for the parse number of days helper.

So I'm not sure I understand the intent here?

PR #156, opened 2025-06-27
Move parsers/validators to validation.py

himdel:
Follows #155

Cosmetic changes only (once rebased after 155), and tests.

Move `parse_number_of_days`, `parse_date_param`, `handle_month` and `_handle_datelike` to `validation.py` where the other validation/parsing functions already are.
(No other changes other than import names.)

Unify the 2 commands to set `.help` and `.help_texts` the same way - and we don't need a constructor for that.

(The next step will be to merge `parse_date_param` (build_report), `handle_datelike` (gather), and maybe `handle_validate_date_param` (validation), so this is just preparing ground for *that*.)

### DIFF
--- a/metrics_utility/automation_controller_billing/helpers.py
+++ b/metrics_utility/automation_controller_billing/helpers.py
@@ -1,20 +1,8 @@
-import datetime
 import json
 
-from datetime import timezone
 from itertools import chain
 
 import pandas as pd
-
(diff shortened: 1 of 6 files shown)


MilanPospisil (approved):
LGTM +1

Provided that rebase will not break anything.

Do not merge before rebase.

PR #162, opened 2025-07-07
Deduplication - ccsp-experimental

himdel:
Epics: AAP-47867, AAP-38205, ANSTRAT-1059 :shrug: 
Issue: AAP-47907, or AAP-47875
(also contains #160 as a first commit now)
(obsoletes #148, but parts of that PR will still be relevant for hostname handling)

This is a weathered-down version of the SDP-0042-P4 proposal for deduplication.
By default this changes nothing.

With `METRICS_UTILITY_DEDUPLICATOR=ccsp-experimental` (along with `METRICS_UTILITY_REPORT_TYPE=CCSPv2`), this will use an experimental deduplication implementation, which still merges everything by `host_name` as before, but then applies additional pass merging hosts by `ansible_product_serial` and `ansible_machine_id` (from `main_host` `canonical_facts`), when both are present and identical.

Unlike the proposal:
  * doesn't handle ports
  * doesn't unify hostname handling
  * keeps hostnames as the main key, does not allow hostname-less dedup algorithms
  * doesn't entierly merge ccsp & renewal guidance logic (a follow-up pr should)
  * does allow to compare first-stage dedup (hostname) with the new dedup (serials) in the same report run
     * but does not expose the info in any reports yet

TODO(wip): unit tests for dedup

---

Structural changes:

* a Dedup step (with its own factory) is added between the DataframeEngineFactory step and the Report building step
* DataframeFactory now returns a dict of named dataframe class instances, not a tuple of built dataframes
* Dedup now builds dataframes, and either returns them as is (dict of named built dataframes), or applies a dedup step
* Reports now don't call dedup, and don't use numerical indexes for dataframes
* `--help` handling shows an `ENVIRONMENT` section
* factories refactored a bit
* dataframe classes now implement an `empty()` - empty dataframe, `group()` group individual entries into a rollup, `regroup()` re-rolls the rollup after hostnames were updated, `operations` just keeps the data for `summarize_merged_dataframe`; `dedup()` can apply a hostname mapping and call regroup

---

Follow-up TODOs:

* enable such an experimental mode in `DedupRenewal`, not only `DedupCCSP` - AAP-49226
  * needs agreement on what to replace it with, since the default renewal dedup already dedups on either machine_id or product_serial, applying an extra step deduping on the compount key does nothing
  * probably replace renewal algo with ccsp algo too, then add the experimental step?

* ansible_host_name vs host_name consistency, deal with ansible_host_port - AAP-49231
  * originally was supposed to be part of the choice of dedup algorithm, but not possible when preserving the original hostname step
  * but we CAN add an extra param to change hostname handling orthogonal to dedup
  * so add one, support the `ansible_host_name || host_name` logic as an option, bare `host_name` too, *and* `host_name:non-default-port` (+- handling of ipv6 addresses with correct port syntax)

* performance testing of new dedup - AAP-49232
  * regression testing without the option
  * what kind of slow-down to expect *with* the option

* consider updating gather to make job_host_summary data also have facts, same as indirect_nodes
  * seems to be a prerequisite to merging indirect with direct based on anything other than hostnames

* docs, more tests

* add all relevant env vars to the ENVIRONMENT section of --help, do the same for gather - AAP-49235

* update the report to add some before and after dedup info (new tab? just two sets of counts in inventory tab?) - AAP-49237


Makefile
```diff
@@ -21,4 +21,10 @@ fix:
 compose:
 	docker compose -f tools/docker/docker-compose.yaml up
 
-.PHONY: help sync test coverage lint fix compose
+clean:
```
MilanPospisil, line 24: :+1:

metrics_utility/automation_controller_billing/dataframe_engine/dataframe_inventory_scope.py
```diff
@@ -2,18 +2,21 @@
...
 
 
 logger = logging.getLogger(__name__)
 
-#######################################
-# Code for building of the dataframe report based on MainHost table
-######################################
 
+def compute_serial(row):
+    facts = parse_json(row['canonical_facts'])
```
MilanPospisil, line 13: Should we also look for facts, or should the facts be primary source of those variables instead of canonical facts?

MilanPospisil (approved):
LGTM :+1: 

However, for next steps, we should discuss:

1) Are we going to search also the facts and not only canonical facts? The variables can be stored in facts instead

2) The algorithm is not transitive, when there is for example:

host1 [Serial1, Serial2]
host2 [Serial2, Serial3]
host3 [Serial4, Serial3]

it will join host1 and host2, but not host3.

We must discuss the improvement of the algorithm.

PR #165, opened 2025-07-10
Adding collector providing the total of vCPU of all workers

itdove:
[AAP-48831](https://issues.redhat.com/browse/AAP-48831)

## Description

- What is being changed? Add a collector for to send the total number of CPU of all workers
- Why is this change needed? To bill per CPU usage rather than flat rate
- How does this change address the issue? Adding a collector

### Steps to Test

1. Deploy an mgt with my own image and the sre branch
2. verify the METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED is false in the mgt config map
3. Deploy a customer instance with my own sre and model branch
4. verify the METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR is true in the new cronjob env and false metric config map read by the existing cronjob.
5. verify the METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED is false in the metric config map
6. Load my own metrics_utility image in the new cronjob.
7. Run the cmd python /metrics_utility/manage.py gather_automation_controller_billing_data --dry-run
8. Check the generated file and it contains vcpu = 1 which is correct.
9. Update the config.env in customer github to set METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED to true
10. (migrate customer from old billing to usage billing scenario)
11. Rerun  python /metrics_utility/manage.py gather_automation_controller_billing_data --dry-run
12. Check the generated file and it contains vcpu = 12 which is correct.
13. The calculation details is displayed
14. Delete the customer instance
15. Change the config.env of the mgt by setting METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED to true, configmap get updated and mgt restarted.
16. Deploy a new customer instance with my own sre and model branch
17. Check the metric configmap and METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED is true and it is correct as the same env var is true in the mgt.
18. Load my own metrics_utility image in the new cronjob.
19. Run the cmd python /metrics_utility/manage.py gather_automation_controller_billing_data --dry-run
20. Check the generated file and it contains vcpu = 12 which is correct.
21. The calculation details is displayed
22. Delete the customer instance
23.
24.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "botocore==1.35.96",
     "distro==1.9.0",
     "django==4.2.16",
+    "kubernetes>=33.1.0",
```
Ladas, line 14: ok, I think these are all dev deps, rou main prod deps are in https://github.com/ansible/metrics-utility/blob/devel/setup.cfg#L13

and then we'll need to talk to PDE to to add this dependency to metrics-utility RPM, which then will get to built image so the operator can use it. (afaik we still built from RPMs, I haven't played with Konflux yet)

nnguyenrh: Thread with PDE started here: https://redhat-internal.slack.com/archives/C01ULNLT6AY/p1752247212704209

metrics_utility/management/validation.py
```diff
@@ -222,18 +223,33 @@ def validate_collectors(errors):
     Environment Variables:
         METRICS_UTILITY_OPTIONAL_COLLECTORS (str, optional): Comma-separated
             list of collector names. Defaults to 'main_jobevent' if not set.
+        METRICS_UTILITY_MANDATORY_COLLECTORS (str, optional): Comma-separated
```
robinbobbitt: @Ladas I think we should reconsider this environment variable. It's confusing to me that we have two environment variables now that essentially both mean "what collectors do you want to run?", and we're calling one of the collector classes "mandatory" yet you can turn them off. Shall we just call it what it is, which is `METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR` and default it to false and set it to true in the SaaS?

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -464,3 +480,56 @@ def main_host_table(since, full_path, until, **kwargs):
...
+    if not cluster_name:
+        logger.error('environment variable METRICS_UTILITY_ANSIBLE_SAAS_CLUSTER_NAME is not set')
+        raise MissingRequiredEnvVar('environment variable METRICS_UTILITY_ANSIBLE_SAAS_CLUSTER_NAME is not set')
+
+    now = datetime.now(timezone.utc)
+
+    info = {'cluster_name': cluster_name, 'timestamp': now.isoformat(), 'nodes': []}
+
+    # If METRICS_UTILITY_VCPU_COUNT_ENABLED is not set or not set to true then it returns 1
+    vcpu_count_enabled_str = os.environ.get('METRICS_UTILITY_VCPU_COUNT_ENABLED')
```
robinbobbitt: It feels unexpected that I have to set `METRICS_UTILITY_VCPU_COUNT_ENABLED` explicitly to True to make the `total_workers_vcpu` collector work correctly, when I've already explicitly included it in the collectors I want. When I see `METRICS_UTILITY_VCPU_COUNT_ENABLED` in the config.env for a repo, I expect it means "enable the cronjob that runs the `total_workers_vcpu` collector", not "explicitly set the vcpu to 1". Is there a downside to just calling it `METRICS_UTILITY_VCPU_COUNT_OVERRIDE` and taking the integer value you want to set it to? wdyt?

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -464,3 +480,56 @@ def main_host_table(since, full_path, until, **kwargs):
...
+    vcpu_count_enabled = False
+    if vcpu_count_enabled_str and (vcpu_count_enabled_str.lower() == 'true'):
+        vcpu_count_enabled = True
+    if not vcpu_count_enabled:
+        return {'cluster_name': info['cluster_name'], 'total_workers_vcpu': '1'}
```
robinbobbitt: We're returning a string `'1'` here but I believe in your real samples, we're returning an integer.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -464,3 +480,56 @@ def main_host_table(since, full_path, until, **kwargs):
...
+        try:
+            kube_config.load_kube_config()
+        except kube_config.ConfigException as e:
+            logger.error(f'Could not configure Kubernetes Python client ERROR: {e}')
+            raise MetricsException(f'Could not configure Kubernetes Python client ERROR: {e}')
+
+    # Create a CoreV1Api client
+    api_instance = client.CoreV1Api()
+
+    nodes = api_instance.list_node()
```
robinbobbitt: It looks like here and below there are opportunities where we might fail silently trying to gather the nodes and then find the CPU value.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -464,3 +480,56 @@ def main_host_table(since, full_path, until, **kwargs):
...
+
+    total_workers_vcpu = 0
+    for node_info in nodes.items:
```
robinbobbitt, line 588: We're not filtering the list of nodes here to just the worker nodes. I believe it doesn't matter for our HCP clusters, as the API doesn't include the control plane nodes in the list. For normal clusters though we'd need to filter out the control plane nodes for this to behave the way we've described. Maybe we just add a comment to this effect, rather than adding an unnecessary filter that might come back to bite us later. wdyt?

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -464,3 +480,56 @@ def main_host_table(since, full_path, until, **kwargs):
...
         table='main_host', query=f'COPY ({query}) TO STDOUT WITH CSV HEADER', path=full_path, prepend_query=yaml_and_json_parsing_functions()
     )
+
+
+@register('total_workers_vcpu', '1.0', format='json', description=_('Total workers vCPU'))
+def total_workers_vcpu(since, full_path, until, **kwargs):
+    if 'total_workers_vcpu' not in get_optional_collectors():
+        return None
+
+    cluster_name = os.environ.get('METRICS_UTILITY_ANSIBLE_SAAS_CLUSTER_NAME')
```
robinbobbitt: I wonder if having "ANSIBLE_SAAS" in here is meaningful? Better to just leave it as `METRICS_UTILITY_CLUSTER_NAME` perhaps? I don't believe other env vars reference the saas.

metrics_utility/management/validation.py
```diff
@@ -222,18 +223,33 @@ def validate_collectors(errors):
...
+            list of collector names. Defaults to 'main_jobevent' if not set.
```
robinbobbitt: wrong default listed

robinbobbitt (changes requested):

metrics_utility/automation_controller_billing/collectors.py
robinbobbitt: Discussed with Dom and Leena. To avoid having to set a variable on all of the existing repo config.envs, we'll keep the behavior Dom recommends but rename to `METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED`. Once all customers are off of the legacy billing, we can retire this variable/behavior.
Existing customers won't have METRICS_UTILITY_VCPU_COUNT_ENABLED at all (or could have it set to False). Once usage-based billing goes live, new customers will have METRICS_UTILITY_VCPU_COUNT_ENABLED. 

We will also support a new config option on the management system telling us whether the new usage based billing is live or not, so that we can start setting METRICS_UTILITY_VCPU_COUNT_ENABLED to True on new instances at the moment PM requests by doing a PR to mgt-prod's config.env vs having to time a full production upgrade.

metrics_utility/management/validation.py
Ladas: @robinbobbitt yes I think METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR might be better approach (because we don't plan to add more mandatory collectors, just make the one mandatory enabled/disabled)

metrics_utility/automation_controller_billing/collectors.py
itdove: ok will add a comment

metrics_utility/management/validation.py
itdove: As @Ladas This will change.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -207,6 +212,14 @@ def yaml_and_json_parsing_functions():
 
 @register('job_host_summary', '1.2', format='csv', description=_('Data for billing'), fnc_slicing=daily_slicing)
 def job_host_summary_table(since, full_path, until, **kwargs):
+    disable_job_host_summary_str = os.environ.get('METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR', 'false')
```
MilanPospisil, line 212: This should be communicated with the team. We have METRICS_UTILITY_OPTIONAL_COLLECTORS for this.

And we must double check that disabling job_host_summary will work. Maybe we have some requirements that it is mandatory.

Im preparing PR that allows to gather 0 days (thus only snapshots). We may try both ways. Setting different value than 4 weeks can be valuable not only in this scenario.

metrics_utility/logger.py
```diff
@@ -5,6 +5,10 @@
 logging.basicConfig(format='%(message)s', level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# This logger will log all message info and up
+logger_info_level = logging.getLogger(__name__)
```
MilanPospisil, line 9: Maybe please do not override our default config.

metrics_utility/automation_controller_billing/collectors.py
MilanPospisil: Edit: We need to keep backwards compatibility. You have it right.

itdove: @MilanPospisil We will have 2 cronjob, the existing one + our own and so to avoid to run twice the job_host_summary_table in our own cronjob we will set METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR to true and set the  METRICS_UTILITY_OPTIONAL_COLLECTORS to 'total_workers_vcpu'
This was recommended by @Ladas 
We need a second cronjob because it has to run every 5 mins and not hourly...

metrics_utility/logger.py
itdove: This was discussed with @himdel , I need to overwritde the logger because whatever the log level is, I want to see that information in the log and so we can see the math for the total vcpu calculation and this in case a customer ask for bill proof.

metrics_utility/automation_controller_billing/collector.py
```diff
@@ -53,7 +54,12 @@ def gather(self, dest=None, subset=None, since=None, until=None, billing_provide
         if not self.is_enabled():
             return None
 
-        with self._pg_advisory_lock('gather_automation_controller_billing_lock', wait=False) as acquired:
+        key = 'gather_automation_controller_billing_lock'
+        suffix = os.environ.get('METRICS_UTILITY_COLLECTOR_LOCK_SUFFIX')
+        if suffix:
+            key = f'gather_automation_controller_billing_{suffix}_lock'
+
+        with self._pg_advisory_lock(key, wait=False) as acquired:
```
MilanPospisil, line 62: How this is intented to be used? Each process will define its own lock? What is the purpose of lock then, maybe the logic should be to avoid locking here completely?

Edit: Ladas told me you want to run CPU collection every 5 minutes, job host summary every 10 minutes. So this makes sense, you want separate lock for those.

Is this true?

itdove: Ladas told me that I need a separate lock to avoid a cronjob blocking the other cronjob. The job_host_summary will continue. to run in the hourly cronjob generated by the operator. The CPU collection will run on a new cronjob every 5 min. The new cronjob will not run the job_host_summary (METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR) but the 2 collections can run concurrently and so we need a separate lock to avoid that one prevent the other one to run
Is this clarifying your question?

MilanPospisil: That confirms it, thanks :-)

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -524,3 +537,57 @@ def main_host_table(since, full_path, until, **kwargs):
     return _copy_table(
         table='main_host', query=f'COPY ({query}) TO STDOUT WITH CSV HEADER', path=full_path, prepend_query=yaml_and_json_parsing_functions()
     )
+
+
+@register('total_workers_vcpu', '1.0', format='json', description=_('Total workers vCPU'), fnc_slicing=limit_slicing)
+def total_workers_vcpu(since, full_path, until, **kwargs):
+    if 'total_workers_vcpu' not in get_optional_collectors():
```
MilanPospisil, line 540: This is generating tarball even if it should not. This may be the cause of failing test yesterday. Tarball is empty. This is come from mistake in json collector. We will repair it in follow up PR by ourselves.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -524,3 +537,57 @@ def main_host_table(since, full_path, until, **kwargs):
...
+        try:
+            kube_config.load_kube_config()
+        except kube_config.ConfigException as e:
+            logger.error(f'Could not configure Kubernetes Python client ERROR: {e}')
+            raise MetricsException(f'Could not configure Kubernetes Python client ERROR: {e}')
+
+    # Create a CoreV1Api client
+    api_instance = client.CoreV1Api()
+    if not api_instance:
+        raise MetricsException('Could get a Kube CoreV1Api client')
```
robinbobbitt:
```suggestion
        raise MetricsException('Could not get a Kube CoreV1Api client')
```

metrics_utility/base/README.md
```diff
@@ -83,6 +83,19 @@ Package is also abstract class. You have to implement basically info for POST re
...
+  - `METRICS_UTILITY_COLLECTOR_LOCK_SUFFIX`: [str] Suffix added to the lock name, must be set in case of multiple cronjobs are running and this to avoid one cronjob to prevent the others to run.
+  - `METRICS_UTILITY_DISABLE_SAVE_LAST_GATHERED_ENTRIES`: [true/false] Some collectors are not time based and thus are not using 'since/until' as they are taking a sample of the state at regular point of time. In that case we don't need to save the last gathered entries. Default false.
+
+### Environment variables for total_workers_vcpu collector:
+
+  - `METRICS_UTILITY_CLUSTER_NAME`: Contains the cluster name which is part of the collection payload.
+  - `METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED`: [true/false] In case of true, the payload will contain the actual number of total vcpu accross all workers otherwise the total will be set to 1.
+
+N.B.: The SaaS solution doesn't have a control place and so all nodes are workers, if this collector is used for another solution then the filtering must be implemented.
```
robinbobbitt:
```suggestion
N.B.: The SaaS solution runs on ROSA HCP so all listed nodes are workers, if this collector is used for another solution then the filtering must be implemented.
```

metrics_utility/test/gather/test_gather_ranges.py
```diff
@@ -91,9 +91,11 @@ def test_only_host_scope(cleanup_glob):
     )
 
     # ensure no other tarballs are present in the directory for current date
-    assert len(glob.glob(f'./metrics_utility/test/test_data/data/{year}/{month}/{day}/*.tar.gz')) == 1
+    # assert len(glob.glob(f'./metrics_utility/test/test_data/data/{year}/{month}/{day}/*.tar.gz')) == 1
```
robinbobbitt, line 94: Did you mean to push the changes in this file?

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -524,3 +537,57 @@ def main_host_table(since, full_path, until, **kwargs):
...
+    now = datetime.now(timezone.utc)
+
+    info = {'cluster_name': cluster_name, 'timestamp': now.isoformat(), 'nodes': []}
+    # If METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED is not set or set to false then it returns 1
+    usage_based_billing_enabled_str = os.environ.get('METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED')
+    usage_based_billing_enabled = False
+    if usage_based_billing_enabled_str and (usage_based_billing_enabled_str.lower() == 'true'):
+        usage_based_billing_enabled = True
+    if not usage_based_billing_enabled:
+        return {'cluster_name': info['cluster_name'], 'total_workers_vcpu': 1}
```
robinbobbitt: Shall we add a log entry in this scenario too for the warm fuzzy that our existing customers are behaving correctly?

metrics_utility/test/gather/test_gather_ranges.py
itdove: yes, this was commented as there is an issue when the tarball is empty... the metrics team will fix it later.

metrics_utility/automation_controller_billing/collectors.py
itdove: ok

robinbobbitt (approved):

MilanPospisil (approved):
LGTM :+1: 

Do not merge it yet, wait for code freeze. This is not needed to be merged ASAP into the new release.

PR #168, opened 2025-07-16
Milan fix gather restriction

MilanPospisil:
[AAP-49645]

## Description

1) Move magic constant literal 4 weeks into constant defined in one place
2) User is informed about changes in the since and until
3) All gathering is printed to the user while the gatherer gathers - previously, it gathered in background without any prints during gathering and then printed all the path to temporary tarballs

Now it is printing what is collecting for each day during collection, so it is more interactive.

## Testing

uv run ./manage.py gather_automation_controller_billing_data --since=5months --ship --force

And test different input, no since and until, since only, until only, both

### Steps to Test

1.
2.
3.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers

### DIFF
--- a/metrics_utility/base/collector.py
+++ b/metrics_utility/base/collector.py
@@ -6,6 +6,7 @@ import os
 import pathlib
 import shutil
 import tempfile
+import time
 
 from abc import abstractmethod
 
@@ -51,7 +52,7 @@ class Collector:
         self.packages = {}
 
         self.last_gathered_entries = None
-        self.logger = logger or logging.getLogger('metrics_utility.base.collector')
+        self.logger = logger or logging.getLogger('awx.main.analytics')
         self.log_level = logging.ERROR if self.collection_type != self.SCHEDULED_COLLECTION else logging.DEBUG
 
         self.tmp_dir = None
@@ -173,16 +174,15 @@ class Collector:
         _now = now()
         original_since = since
         original_until = until
-        print('\n')
-        print(f'Original since-until: {original_since} to {original_until}')
+        self.logger.info(f'Original since-until: {original_since} to {original_until}')
 
         # Make sure that the endpoints are not in the future.
         if until is not None and until > _now:
             until = _now
-            print(f'End of the collection interval is in the future, setting to {_now}.')
+            self.logger.info(f'End of the collection interval is in the future, setting to {_now}.')
         if since is not None and since > _now:
             since = _now
-            print(f'Start of the collection interval is in the future, setting to {_now}.')
+            self.logger.info(f'Start of the collection interval is in the future, setting to {_now}.')
 
         # The value of `until` needs to be concrete, so resolve it.  If it wasn't passed in,
         # set it to `now`, but only if that isn't more than 4 weeks ahead of a passed-in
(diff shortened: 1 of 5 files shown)


metrics_utility/management/commands/gather_automation_controller_billing_data.py
```diff
@@ -73,10 +73,7 @@ def handle(self, *args, **options):
         )
 
         tgzfiles = collector.gather(since=since, until=until, billing_provider_params=billing_provider_params)
-        if tgzfiles:
-            for tgz in tgzfiles:
-                self.logger.info(tgz)
-        else:
+        if not tgzfiles:
             self.logger.error('No analytics collected')
```
Ladas, line 77: keep the condition and on success add

 self.logger.info('Analytics collected')

metrics_utility/base/collector.py
```diff
@@ -273,7 +285,26 @@ def _gather_csv_collections(self):
...
+        seconds = 20
+        last_print_time = time.time()
+
+        self.logger.info('Starting collecting, this may take while.')
+
         for collection in self.collections[Collection.COLLECTION_TYPE_CSV]:
+            # print this message every X seconds
+            current_time = time.time()
+
+            if current_time - last_print_time >= seconds:
```
Ladas: hm, this is a fixed list of collections from the collections file? So no need to do any time checks? (this will jsut print ~1-7 lines based on number of collections allowed, or what does this print?)

metrics_utility/management/commands/gather_automation_controller_billing_data.py
Ladas: if you can it's better to:

if <positive condition>:
else:

And ruff should be getting if you're doing if if instead of if else

metrics_utility/base/collector.py
```diff
@@ -169,50 +171,59 @@ def delete_tarballs(self):
...
         _now = now()
+        original_since = since
+        original_until = until
+        self.logger.info(f'Original since-until: {original_since} to {original_until}')
 
         # Make sure that the endpoints are not in the future.
         if until is not None and until > _now:
             until = _now
-            self.logger.warning(f'End of the collection interval is in the future, setting to {_now}.')
+            self.logger.info(f'End of the collection interval is in the future, setting to {_now}.')
```
himdel: these should still be warnings, right?

When we override what the user has set, that's a warning, not an info message

MilanPospisil: I thought info is more what I felt the message should be, but I can change it back, no problem.

metrics_utility/base/collector.py
```diff
@@ -273,7 +284,14 @@ def _gather_csv_collections(self):
          1) the temp file needs to be deleted to ensure enough disk space
          2) Collections with slicing function can produce duplicate filename
         """
+
+        last_key = None
+
         for collection in self.collections[Collection.COLLECTION_TYPE_CSV]:
+            if last_key != collection.key:
+                self.logger.warning(f'Progress info: Now gathering {collection.key}')
```
himdel, line 292: "progress info" suggests something like `.info` more than `.warning`?

metrics_utility/base/collector.py
```diff
@@ -169,6 +171,9 @@ def delete_tarballs(self):
     #
     def _calculate_collection_interval(self, since, until):
         _now = now()
+        original_since = since
+        original_until = until
+        self.logger.warning(f'Original since-until: {original_since} to {original_until}')
```
himdel, line 176: (also more like debug or info?)

metrics_utility/base/collector.py
```diff
@@ -199,20 +206,24 @@ def _calculate_collection_interval(self, since, until):
...
         if last_gather < horizon:
             last_gather = horizon
-            self.logger.warning(f'Last analytics run was more than 4 weeks prior to {until}, using {horizon} instead.')
+            self.logger.warning(f'Last analytics run was more than {self.MAX_GATHER_PERIOD_WEEKS} weeks prior to {until}, using {horizon} instead.')
 
         self.gather_since = since
         self.gather_until = until
         self.last_gather = last_gather
 
+        self.logger.warning(f'Final since-until: {since} to {until}')
```
himdel, line 225: (also more like debug or info?)

himdel (approved):
LGTM :)

Some of the logger.warnings could perhaps be something other than warning, but the MAX_GATHER_PERIOD_WEEKS change works :)

PR #171, opened 2025-07-18
[AAP-49226] Add new deduplication modes and corresponding tests for renewal guidance

cshiels-ie:
- Introduced DedupRenewalHostname and DedupRenewalExperimental classes for enhanced deduplication strategies based on hostname and experimental logic.
- Updated Factory class to support new deduplication modes: 'renewal-hostname' and 'renewal-experimental'.
- Implemented unit tests for new classes and factory methods, ensuring correct initialization and functionality.
- Enhanced existing tests for DedupRenewal to cover new deduplication scenarios and edge cases.
- Added comprehensive test cases for the new deduplication strategies, validating their behavior with sample data.

Available Options for RENEWAL_GUIDANCE
You can explicitly set METRICS_UTILITY_DEDUPLICATOR to:

1. 'renewal' → Standard Deduplication used already and the default option. 
2. 'renewal-hostname' → DedupRenewalHostname (simpler hostname-only deduplication, similar to CCSP)
3. 'renewal-experimental' → DedupRenewalExperimental (combines hostname + serial-based deduplication)


metrics_utility/automation_controller_billing/dedup/factory.py
```diff
@@ -17,22 +21,36 @@ def create(self):
...
+            return DedupCCSP(**kwargs)
+        if report_type in {'RENEWAL_GUIDANCE'}:
+            return DedupRenewal(**kwargs)
+        raise NotSupportedFactory(f'Unknown report type: {report_type}')
 
+    def _get_custom_deduplicator(self, deduplicator, report_type, kwargs):
         if deduplicator == 'ccsp':
             return DedupCCSP(**kwargs)
         if deduplicator == 'renewal':
             return DedupRenewal(**kwargs)
```
AlexSCorey, line 42: Do we need this code still if we are introducing _get_default_deduplicator()? https://github.com/ansible/metrics-utility/pull/171/files#diff-e2af984d41b5bf63f7d18d962086fb91bee92aea8cfeba988fd874ef80fd7dc6R31

metrics_utility/automation_controller_billing/dedup/renewal_guidance.py
```diff
@@ -77,3 +77,221 @@ def find_dupes(self, dupes, key, values):
...
+                    'last_automation': max(dupes['last_automation']),
+                    'automated_counter': sum(dupes['automated_counter']),
+                    'deleted_counter': sum(dupes['deleted_counter']),
+                    'last_deleted': max(dupes['last_deleted']),
+                }
+            )
+
+        return {'host_metric': pd.DataFrame(deduped_list)}
+
+    def stringify(self, value):
```
AlexSCorey: We have this stringify and a function by the same name on line 72 of this same file.  Can we combine them, and maybe put them into a utility file somewhere?

himdel (approved):
:+1: Not seeing any problems, apart from failing tests and what sonarcloud says :)

metrics_utility/automation_controller_billing/dedup/factory.py
cshiels-ie: We don't really need it but at the same time there is no harm having it here as it helps with the readability of the code and it has no effect on the performance.

metrics_utility/automation_controller_billing/dedup/renewal_guidance.py
cshiels-ie: I refactored to fix this issue thanks 👍

PR #172, opened 2025-07-22
 [AAP-48952] Added in Try Catch to the main_indirectmanagednodeaudit_table collector

cshiels-ie:
## Description
Links to: [AAP-48952](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=21043&projectKey=AAP&view=detail&selectedIssue=AAP-48952)

### What is being changed?
A try...except block has been added to the data collection method for the main_indirectmanagednodeaudit_table. This change specifically handles the ProgrammingError that occurs when this table is not present in the database. Additionally, new integration tests have been added.

### Why is this change needed?
 This change is required to safely backport the enhanced deduplication functionality to the 2.4 release stream. The database schema for version 2.4 does not include the main_indirectmanagednodeaudit_table. Without this fix, running the metrics collector against a 2.4 environment would result in a crash, preventing any metrics from being gathered.

### How does this change address the issue?
By catching the specific database error, the collector can now gracefully handle the absence of the table. Instead of failing, it logs a warning message, returns None for the missing data, and allows the deduplication process to proceed with the tables that are available. This ensures backward compatibility and allows the collector to function correctly on older release streams while still providing enhanced functionality on newer versions.

## Testing

- Access to an environment with a database schema corresponding to the 2.4 release (i.e., without the main_indirectmanagednodeaudit_table).
- Access to an environment with a more recent database schema where the table exists.

## Steps to Test

- Execute the ansible-metrics collector against a database where main_indirectmanagednodeaudit_table is present.
- Execute the ansible-metrics collector against a 2.4 database where main_indirectmanagednodeaudit_table is missing.
- Check the collector logs for the run against the 2.4 database.
- Run the automated test suite, including the newly added integration tests.

## Expected Results

- The collector should run successfully in both scenarios without crashing.
- When run against the 2.4 database, a warning message should be present in the logs indicating that main_indirectmanagednodeaudit_table was not found and that the collector is proceeding.
- All existing and new tests should pass successfully, validating both the new and existing functionality.

## Notes for Reviewers

The primary goal of this PR is to ensure backward compatibility for the 2.4 release stream. The core change is the simple try...except block that prevents the collector from crashing on older database schemas. The new tests specifically mock the ProgrammingError to confirm the graceful failure mechanism works as intended.


himdel (approved):
Sounds good, should gracefully handle the missing table in 2.4

PR #173, opened 2025-07-22
logger - disable debug logs (without --verbose), unify logger

himdel:
The debug log is too noisy to enable by default. (https://github.com/ansible/metrics-utility/actions/runs/16442073256/job/46465377532?pr=171#step:13:100680)

Previously this was manageble since this only affected the awx.main.analytics logger but since #168, all the loggers are `awx.main.analytics` now.

Making sure the logger now defaults to info and above, and only outputs debug when --verbose.
(We may need to switch to the default warning level if it's still too much, but trying the first step first.)

---

Issue: AAP-49802

Remove `print` statements, either remove or replace with `logger.info` or `logger.debug`.
(Kept `print` in `testathon/` since none of that code is using loggers now.)

Remove `getLogger('awx.main.analytics')`, don't reuse a logger config, have our own.

All logging is now configured in `metrics_utlity.logger`, defaults to outputting info, warnings & errors now,
add debug when `--verbose`.


AlexSCorey (commented):
This looks good.  I like the idea.  I wonder if we can write a lint rule...or some other kind of rule to only allow use of metrics_utility.logger?

MilanPospisil (approved):
LGTM :+1: 

Tested, works, makes sense. Thanks Martin!

himdel:
Thanks! :)

@AlexSCorey good idea with the linter .. added one in a8684f1a9e7cb576af6d24c8fe51f42e8b49ffae

(it will output a list of files that are using `logging.getLogger` in metrics_utility/, except for logger.py - and if nonempty, output a message and fail the check)

PR #174, opened 2025-07-22
Milan testathon generate

MilanPospisil:
[AAP-49142]

## Description

Description of scripts in readme.

It inserts inventories, hosts, job templates, run them, then modify jobhostsummary data to have multiple date ranges available.

Gather all script then calling gather in loop for each month.

It can connect to local environment (containerized) and to remote RPM. Other server types will be added later.

## Testing

Locally:
run testathon/testathon_data_prepare.py

You do need to set anything if your app is using defaults (readme)

then run gather_all.py

This also do not need any env variables.

RPM:
you must set variables described in readme.
This script will then log into RPM instance and start adding data there.

### Steps to Test

1.
2.
3.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers


testathon/testathon_data_prepare.py
```diff
@@ -0,0 +1,393 @@
...
+        if Failed:
+            print(f'Failed to delete job template {id}, maybe jobs are still running, trying again in 10 seconds')
+            time.sleep(10)
+        else:
+            break
+
+
+def search_for_demo_project():
+    # https://localhost:8030/api/controller/v2/projects/?search=Demo+Project&order_by=name&page=1&page_size=10
+    url = f'{API_URL}/projects/?search=Demo+Project&order_by=name&page=1&page_size=10'
```
AlexSCorey: We shouldn't assume that the Demo Project exists.  While it is provided out of the box, it can be deleted.

testathon/readme.md
```diff
@@ -0,0 +1,34 @@
+# testathon_data_prepare
+
+if you run this directly without any parameters, it will fill local environment db, provided that it runs
```
AlexSCorey:
```suggestion
If you run this directly without any parameters, it will pre-populate local environment db, provided that it runs
```

testathon/readme.md
```diff
@@ -0,0 +1,34 @@
+# testathon_data_prepare
```
AlexSCorey, line 1: Lets add some example commands to this doc to add clarity for how to use this tool.

AlexSCorey (commented):
I think this is a really good start.  I've left a few comments/questions.  Let me know if you have questions.

testathon/gather_all.py
```diff
@@ -0,0 +1,129 @@
...
+            f'--since={since.isoformat()}',
+            f'--until={until.isoformat()}',
+            '--force',
+        ]
+
+        if SSH_URL and SSH_USER:
+            # Build remote SSH command list
+            env_list = [f'{k}={v}' for k, v in config.items()]
+            ssh_cmd = ['ssh', f'{SSH_USER}@{SSH_URL}', 'sudo', '-E', 'env', *env_list, 'metrics-utility', *args]
+            print('Running remote:', ' '.join(ssh_cmd))
```
himdel, line 87: This is prone to breaking with interesting values ... definitely won't do the right thing with  `METRICS_UTILITY_REPORT_SKU_DESCRIPTION`.

`f'{k}="{v}"'` is probably slightly better, as long as users aren't allowed to change things, or you'll have to escape `v` before quoting

testathon/gather_all.py
```diff
@@ -0,0 +1,129 @@
...
+            result = subprocess.run(ssh_cmd, check=False, capture_output=True, text=True)
+        else:
+            # Local docker exec path
+            docker_env = []
+            for k, v in config.items():
+                docker_env += ['-e', f'{k}={v}']
```
himdel, line 93: same here

(but if you're running a local command, you can just use `os.execve` and not deal with any of the excaping nonsense)

testathon/gather_all.py
MilanPospisil: This description should not be there at the first place. I copied it from metrics-utility/run-ccsp2-gather example without thinking about it, so I will delete it from there to.

Gather only needs:
export METRICS_UTILITY_SHIP_PATH
export METRICS_UTILITY_SHIP_TARGET

testathon/gather_all.py
MilanPospisil: Thanks, its not necessary now with only two env variables left and they are not set by user, so we know, they work.

testathon/testathon_data_prepare.py
```diff
@@ -0,0 +1,441 @@
...
+                break
+        except Exception as e:
+            print(f'Failed to delete job template {id}: {e}')
+            Failed = True
+
+        if Failed:
+            print(f'Failed to delete job template {id}, maybe jobs are still running, trying again in 10 seconds')
+            time.sleep(10)
+        else:
+            break
```
himdel:
```suggestion
    failed = True
    while True:
        try:
            url = f'{API_URL}/job_templates/{id}/'
            resp = requests.delete(url, auth=(USERNAME, PASSWORD), verify=VERIFY_SSL)
            resp.raise_for_status()
            print(f'Deleted job template {id}')
            if resp.status_code not in (200, 202, 204):
                print(f'Failed to delete job template {id}: {resp.status_code} - {resp.text}')
            else:
                failed = False
                break
        except Exception as e:
            print(f'Failed to delete job template {id}: {e}')
            failed = True

        if failed:
            print(f'Failed to delete job template {id}, maybe jobs are still running, trying again in 10 seconds')
            time.sleep(10)
        else:
            break
```

only Classnames should start with a capital

MilanPospisil: Oooh, good catch!

himdel (approved):
LGTM, not seeing anything else :)

PR #178, opened 2025-07-25
Fix for METRICS_UTILITY_OPTIONAL_COLLECTORS being set as a whitespace 

cshiels-ie:
Fixing an issue that blocked the option to have no optional collectors


metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -36,7 +36,10 @@ def something(since):
 
 
 def get_optional_collectors():
-    return os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent').split(',')
+    opt_collector = os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent')
+    if not opt_collector or opt_collector.isspace():
+        opt_collector = 'main_jobevent'
+    return opt_collector.split(',')
```
himdel: With `METRICS_UTILITY_OPTIONAL_COLLECTORS=""` this should return an empty array, not `['main_jobevent']` (and not `['']`)
(The default should only kick in with `unset METRICS_UTILITY_OPTIONAL_COLLECTORS`.)

(ie it should still be possible to disable the default optional collector, without adding any others)

metrics_utility/management/validation.py
```diff
@@ -227,7 +227,10 @@ def validate_collectors(errors):
         - Error messages include the invalid collector names and the list of
           valid values.
     """
-    collectors = os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent').split(',')
+    collectors = os.environ.get('METRICS_UTILITY_OPTIONAL_COLLECTORS', 'main_jobevent')
+    if not collectors or collectors.isspace():
```
itdove: Can't you call the get_optional_collectors rather than duplicate the code?

itdove: you might have a cycle between package....

metrics_utility/automation_controller_billing/collectors.py
cshiels-ie: I rolled back the collector as the change was just needed in the validator

itdove (approved):

himdel (approved):
Yeah, that should work too :D :+1:

PR #184, opened 2025-08-01
AAP-50701 - Fill Data for Testathon 2

MilanPospisil:
[[AAP-50701](https://issues.redhat.com/browse/AAP-50701)]

## Description

Allows pushing data into containerized and OCP.

Code to access those environments was added.

## Testing

You can use builds there:
https://jenkins-csb-aap-main.dno.corp.redhat.com/job/AAPQA/job/Sandbox/job/abakshi/job/instances_milan/

But they may be inactive by time of testing, so you will have to rerun them.

More instructions in:
testathon/readme.md

### DIFF
--- a/testathon/gather_all.py
+++ b/testathon/gather_all.py
@@ -6,13 +6,15 @@ import sys
 
 from datetime import date, timedelta
 
+from helper_ocp_prepare import create_oc_environs
+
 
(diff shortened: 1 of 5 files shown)


himdel (approved):
LGTM :)

(The ssh key won't always be `id_ed19561` but I'd assume people can figure out which key they're using.)

PR #185, opened 2025-08-06
Add dest path in --verbose for gather

MilanPospisil:
[AAP-[50977](https://issues.redhat.com/browse/AAP-50977)]

## Description

Add small change for better logging the info:

1) with --verbose on, it will log path of stored tarballs
2) when logging start of collecting of collector types, it checks if the collector is on, otherwise it logs that collector is disabled

## Testing
METRICS_UTILITY_SHIP_PATH=/var/tmp/shipped_data METRICS_UTILITY_SHIP_TARGET=directory uv run ./manage.py gather_automation_controller_billing_data --ship --since=2024-02-29 --until=2024-02-29 --force

METRICS_UTILITY_SHIP_PATH=/var/tmp/shipped_data METRICS_UTILITY_SHIP_TARGET=directory uv run ./manage.py gather_automation_controller_billing_data --ship --since=2024-02-29 --until=2024-02-29 --force --verbose

### DIFF
--- a/metrics_utility/automation_controller_billing/package/package_directory.py
+++ b/metrics_utility/automation_controller_billing/package/package_directory.py
@@ -1,6 +1,5 @@
 import os
 import shutil
-import sys
 
 from django.conf import settings
 
@@ -62,9 +61,7 @@ class PackageDirectory(base.Package):
         os.makedirs(os.path.dirname(destination_path), exist_ok=True)
         shutil.copyfile(self.tar_path, destination_path)
 
-        # if args contains --verbose, print the destination path
-        if '--verbose' in sys.argv:
-            logger.info(f'tarball saved to: {destination_path}')
+        logger.debug(f'tarball saved to: {destination_path}')
 
         self.shipping_successful = True
         return True
--- a/metrics_utility/automation_controller_billing/package/package_directory.py
+++ b/metrics_utility/automation_controller_billing/package/package_directory.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+import sys
 
 from django.conf import settings
 
@@ -60,6 +61,10 @@ class PackageDirectory(base.Package):
 
         os.makedirs(os.path.dirname(destination_path), exist_ok=True)
         shutil.copyfile(self.tar_path, destination_path)
+
+        # if args contains --verbose, print the destination path
+        if '--verbose' in sys.argv:
+            logger.info(f'tarball saved to: {destination_path}')
 
         self.shipping_successful = True
         return True
--- a/metrics_utility/automation_controller_billing/package/package_s3.py
+++ b/metrics_utility/automation_controller_billing/package/package_s3.py
@@ -1,4 +1,5 @@
 import os
+import sys
 
 from django.conf import settings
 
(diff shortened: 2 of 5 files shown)


metrics_utility/automation_controller_billing/package/package_directory.py
```diff
@@ -61,5 +62,9 @@ def ship(self):
         os.makedirs(os.path.dirname(destination_path), exist_ok=True)
         shutil.copyfile(self.tar_path, destination_path)
 
+        # if args contains --verbose, print the destination path
+        if '--verbose' in sys.argv:
+            logger.info(f'tarball saved to: {destination_path}')
```
himdel:
```suggestion
        logger.debug(f'tarball saved to: {destination_path}')
```

logger.debug is only enabled with --verbose .. you don't need to put if --verbose everywhere

metrics_utility/automation_controller_billing/package/package_directory.py
```diff
@@ -61,5 +62,9 @@ def ship(self):
         os.makedirs(os.path.dirname(destination_path), exist_ok=True)
         shutil.copyfile(self.tar_path, destination_path)
 
+        # if args contains --verbose, print the destination path
```
himdel:
```suggestion
```

metrics_utility/automation_controller_billing/package/package_s3.py
```diff
@@ -61,5 +62,9 @@ def ship(self):
         s3_handler = S3Handler(params=self.collector.billing_provider_params)
         s3_handler.upload_file(self.tar_path, object_name=destination_path)
 
+        # if args contains --verbose, print the destination path
+        if '--verbose' in sys.argv:
+            logger.info(f'tarball saved to: {destination_path}')
```
himdel:
```suggestion
        logger.debug(f'tarball saved to: {destination_path}')
```

metrics_utility/base/collector.py
```diff
@@ -289,9 +289,26 @@ def _gather_csv_collections(self):
...
+                if collection.key == 'job_host_summary' and not disable_job_host_summary:
+                    write_enabled = True
+
+                if collection.key in optional_collectors:
+                    write_enabled = True
+
+                if write_enabled:
+                    logger.warning(f'Progress info: Now gathering {collection.key}')
+                else:
+                    logger.warning(f'Progress info: Skipping {collection.key} because it is not enabled.')
```
himdel, line 310:
```suggestion
                if write_enabled:
                    logger.debug(f'Progress info: Now gathering {collection.key}')
                else:
                    logger.debug(f'Progress info: Skipping {collection.key} because it is not enabled.')
```

(unless you intentionally want to list every skipped collector on every collect?, without verbose?)

MilanPospisil: This should be maybe info, but previously, it was part of standard output without --verbose.

It prints every collector only once, but current change is informing user that this collector is skipped.

himdel: Aah, makes sense :+1: :)

metrics_utility/base/collector.py
```diff
@@ -289,9 +289,26 @@ def _gather_csv_collections(self):
 
         last_key = None
 
+        disable_job_host_summary_str = os.environ.get('METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR', 'false')
+        disable_job_host_summary = disable_job_host_summary_str.lower() == 'true'
+
+        optional_collectors = get_optional_collectors()
+
         for collection in self.collections[Collection.COLLECTION_TYPE_CSV]:
```
himdel, line 297: is this something we'd want to do for JSON type collectors too?

(if not, lgtm, but seems like we're only doing this for csv here)

himdel (approved):
LGTM :) (but see question re csv vs csv & json)

PR #186, opened 2025-08-06
Document environment variables, extra_params

himdel:
Document which env variables are used where,
and same for extra_params / billing_provider_params.

---

For the service, it will be important that the command classes read all the env vars (except for AWX_PATH),
and provide a validated processed version of these to the worker classes.

Also, the extra params should be split into, like, s3 params, collector params, report params, and only passed to the right places. That way, these classes can have well-defined small interfaces callable from metrics-service workers.

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -1,0 +1,243 @@
+All known metrics-utility environment variables:
+
+```
+AWX_PATH
+KUBERNETES_SERVICE_PORT
+METRICS_UTILITY_BILLING_ACCOUNT_ID
+METRICS_UTILITY_BILLING_PROVIDER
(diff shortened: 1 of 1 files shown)


himdel:
`METRICS_UTILITY_USAGE_BASED_BILLING_ENABLED` -> `METRICS_UTILITY_USAGE_BASED_METERING_ENABLED` (198)
\+ `METRICS_UTILITY_PROMETHEUS_URL` (198)

himdel:
cursor review

MilanPospisil (approved):
LGTM :+1

Validated locally with my cursor agent and it makes sense :-) Thanks, its useful :)

PR #187, opened 2025-08-06
Code cleanup

himdel:
* Use `os.getenv`, not `os.environ.get`; drop default if None (consistency only)
* Cleanup more dead code/props
* Change `f'{val}'` to `val` or `str(val)`
* Separate `get_install_type` from `config` collector
* No `price_per_node` in renewal guidance
* `build_report`: use more readable variables; use `date.today` instead of `datetime.now`; move extractor to after `extra_params` is fully set
* `validate_ship_path`: don't mutate global const, remove unused gather code path

### DIFF
--- a/metrics_utility/management/commands/build_report.py
+++ b/metrics_utility/management/commands/build_report.py
@@ -99,14 +99,14 @@ class Command(BaseCommand):
         opt_ephemeral = parse_number_of_days(options.get('ephemeral'))
         opt_force = options.get('force')
 
-        ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET', None)
+        ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET')
         extra_params = self._handle_extra_params(ship_target)
         extra_params['opt_since'] = opt_since
         extra_params['opt_until'] = opt_until
         extra_params['ephemeral_days'] = opt_ephemeral
         extra_params['month_since'] = month
         extra_params['month_until'] = next_month
-        extra_params['deduplicator'] = os.getenv('METRICS_UTILITY_DEDUPLICATOR', None) or None
+        extra_params['deduplicator'] = os.getenv('METRICS_UTILITY_DEDUPLICATOR') or None
 
         extractor = ExtractorFactory(ship_target, extra_params).create()
 
@@ -174,7 +174,7 @@ class Command(BaseCommand):
     def _handle_extra_params(self, ship_target=None):
         base = self._handle_ship_target(ship_target)
 
-        report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE', None)
+        report_type = os.getenv('METRICS_UTILITY_REPORT_TYPE')
         price_per_node = float(os.getenv('METRICS_UTILITY_PRICE_PER_NODE', 0))
 
         if not report_type:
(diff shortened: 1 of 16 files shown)


metrics_utility/management/commands/build_report.py
```diff
@@ -99,14 +99,14 @@ def handle(self, *args, **options):
...
-        ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET', None)
+        ship_target = os.getenv('METRICS_UTILITY_SHIP_TARGET')
         extra_params = self._handle_extra_params(ship_target)
         extra_params['opt_since'] = opt_since
         extra_params['opt_until'] = opt_until
         extra_params['ephemeral_days'] = opt_ephemeral
         extra_params['month_since'] = month
         extra_params['month_until'] = next_month
-        extra_params['deduplicator'] = os.getenv('METRICS_UTILITY_DEDUPLICATOR', None) or None
+        extra_params['deduplicator'] = os.getenv('METRICS_UTILITY_DEDUPLICATOR') or None
```
MilanPospisil, line 111: Is it even necessary? This should return None by default if variable not found.

Habbit from javascript? :-)

MilanPospisil:
Needs rebase.

metrics_utility/management/commands/build_report.py
```diff
@@ -109,24 +110,26 @@ def handle(self, *args, **options):
...
-
         # Determine destination path for generated report and skip processing if it exists
+        report_type = extra_params['report_type']
+        ship_path = extra_params['ship_path']
         if opt_since is not None:
-            now = datetime.datetime.now().replace(second=0, microsecond=0, tzinfo=timezone.utc)
-            extra_params['since_date'] = opt_since.date()
-            extra_params['until_date'] = opt_until.date() if opt_until else now.date()
+            since_date = opt_since.date()
+            until_date = opt_until.date() if opt_until else datetime.date.today()
```
MilanPospisil, line 118: This is practicaly removing datetime granularity and setting date granularity.

Are you sure we will not need it in service? I can imagine we will be building reports for hours.

metrics_utility/management/validation.py
```diff
@@ -276,16 +276,13 @@ def validate_ship_path(errors, ship_target, method):
         - For 'directory' ship target, checks if METRICS_UTILITY_SHIP_PATH is an existing directory.
         - Appends an error message to 'errors' if the directory does not exist.
     """
-    no_path = 'No Path Provided'
-    ship_path = os.getenv('METRICS_UTILITY_SHIP_PATH', no_path)
-    dir_paths = VALID_SHIP_TARGET_BUILD
-    if 's3' in dir_paths:
-        dir_paths.remove('s3')
```
MilanPospisil, line 284: Oh, this mutates original VALID_SHIP_TARGET_BUILD, nice catch!

metrics_utility/management/commands/build_report.py
himdel: If not found, yeah, but `METRICS_UTILITY_DEDUPLICATOR= command` will find an empty string, so this just turns `""` into `None`

MilanPospisil: This is evil...

MilanPospisil (approved):
LGTM :+1:

PR #188, opened 2025-08-06
pyproject: add more data, use dynamic version

himdel:
Related to AAP-54478

RPM builds still delete pyproject and use setup.cfg.

But any library builds for pypi will use pyproject and not setup.

Adding more package details to pyproject, and switching to consume dynamic version from `setup.cfg`.
Removes any references to 0.5.1dev (current future version) except for changelog and setup.cfg (and uv.lock).

(Tested on containerized, returns '0.5.1' (`/var/lib/awx/env/bin/python`; `from importlib.metadata import version`; `version('metrics-utility')`))

### DIFF
--- a/metrics_utility/automation_controller_billing/collectors.py
+++ b/metrics_utility/automation_controller_billing/collectors.py
@@ -4,6 +4,7 @@ import os.path
 import platform
 
 from datetime import datetime, timezone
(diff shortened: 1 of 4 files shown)


cshiels-ie (approved):
LGTM

PR #189, opened 2025-08-07
When json collector is disabled (returns None), it will generate tarball even it should not

MilanPospisil:
[AAP-[50991](https://issues.redhat.com/browse/AAP-50991)]

## Description

On devel - tarball is generated for json collector that is not empty, which is wrong. Tarball should not be generated, just as it works for csv collectors.

This PR checks for empty data in json collectors and not store tarball at all.

## Testing

Add this into:
metrics_utility/automation_controller_billing/collectors.py

@register('test_json', '1.0', format='json', description=_('JSON'), fnc_slicing=limit_slicing)
def test_json(since, full_path, until, **kwargs):
    return None
    
This will create empty json collector. Now run:

METRICS_UTILITY_SHIP_PATH=/var/tmp/shipped_data METRICS_UTILITY_SHIP_TARGET=directory uv run ./manage.py gather_automation_controller_billing_data --ship --since=2024-02-29 --until=2024-02-29 --force

You should not see any tarball with json (containing no data).
Without this PR on devel, tarball should be generated.

### DIFF
--- a/metrics_utility/automation_controller_billing/collectors.py
+++ b/metrics_utility/automation_controller_billing/collectors.py
@@ -429,6 +429,11 @@ def main_indirectmanagednodeaudit_table(since, full_path, until, **kwargs):
         return None
 
 
+@register('test_json', '1.0', format='json', description=_('JSON'), fnc_slicing=limit_slicing)
+def test_json(since, full_path, until, **kwargs):
+    return None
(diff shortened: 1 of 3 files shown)


cshiels-ie (approved):
LGTM 👍

PR #191, opened 2025-08-14
Generator & perf test updates

himdel:
Issue: AAP-49232 / AAP-51583

Moves performance testing related code under `tools/perf/`.
Moves the generated data from `metrics_utility/test/test_data/data/2026` to `tools/perf/generated/data/2026`.
gitignore `tools/perf/generated/`.
Drop attempt to generate 10M, 1h of generating time and 7-10 hours of build report time is too much and hits xlsx limits.
Add a readme for the perf scripts.
Add a script to extract copy-pastable timings table from the script output.
Adds missing optional sheets to the all sheets build run, also enables ccsp-experimental dedup (but not for default run).
The generator now generates plausible product_serial and machine_id values, about 1/16th of hosts should share the same values.

How to run:

```
# run once
tools/perf/run-perf-gen

# run for every tested branch
tools/perf/run-perf-build | tee tmp
tools/perf/extract-timings tmp
```

outputs something like

```
rows	variant	memory	time
10000	default	161792K	26.09s
10000	all	162524K	29.58s
100000	default	514496K	214.53s
100000	all	544204K	228.76s
```


tools/perf/run-perf-build
```diff
@@ -0,0 +1,56 @@
...
+export METRICS_UTILITY_SHIP_TARGET="directory"
+
+run_default_sheets() {
+  unset METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS
+  export METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS
+  unset METRICS_UTILITY_DEDUPLICATOR
+  export METRICS_UTILITY_DEDUPLICATOR
+
+  log "default sheets"
+  measure uv run ./manage.py build_report --force "$@"
```
himdel:
```suggestion
  measure uv run ../../manage.py build_report --force "$@"
```

tools/perf/run-perf-build
```diff
@@ -0,0 +1,56 @@
...
+}
+
+run_all_sheets() {
+  export METRICS_UTILITY_OPTIONAL_CCSP_REPORT_SHEETS="ccsp_summary,data_collection_status,indirectly_managed_nodes,infrastructure_summary,inventory_scope,jobs,managed_nodes,managed_nodes_by_organizations,usage_by_collections,usage_by_modules,usage_by_organizations,usage_by_roles"
+  export METRICS_UTILITY_DEDUPLICATOR="ccsp-experimental"
+
+  log "all sheets"
+  measure uv run ./manage.py build_report --force "$@"
```
himdel:
```suggestion
  measure uv run ../../manage.py build_report --force "$@"
```

tools/perf/README.md
```diff
@@ -0,0 +1,62 @@
+# measuring metrics-utility build\_report performance
+
+* `generator.py` - loads a tarball, duplicates and randomizes data, and saves to new tarball
```
MilanPospisil: Maybe put this at the end of the main list of tools and note that it is wrapped by run-perf-gen.

tools/perf/README.md
```diff
@@ -0,0 +1,62 @@
...
+* `run-perf-gen` - runs the generator script to generate a 3 sets of 10k, 100k & 1M entries
```
MilanPospisil, line 3: Maybe comment the estimated time of execution, Im trying to run this and I do not know if it is even feasible to wait :)

himdel: It's about 2.5 hours on real hardware, probably still less than 10 in a VM , will do :)

EDIT: added :+1:

tools/perf/README.md
himdel: Sorted by usage order :+1:

tools/perf/extract-timings
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/perl
```
MilanPospisil: By the way, why perl? Will the code in python be so much more complicated?
 
 I know its only nitpicking, but since many people are not knowing perl, python would be better. But Im fine with perl - especially with LLMs, short codes in different language can be very easily explained or transformed.

MilanPospisil: For example chatgpt wrote me this (I did not tested it, just for example that in python, it would not be too longer):

```python
import re
import sys

count = 0
variant = ""
mem = ""
time = ""

print("rows\tvariant\tmemory\ttime")

for line in sys.stdin:
    line = line.strip()

    m = re.search(r": build ([\d,]+)$", line)
    if m:
        count = int(m.group(1).replace(",", ""))

    m = re.search(r": (\w+) sheets$", line)
    if m:
        variant = m.group(1)

    m = re.match(r"^memory (\d+K)$", line)
    if m:
        mem = m.group(1)

    m = re.match(r"^time ([\d.]+s)$", line)
    if m:
        time = m.group(1)

    if mem and time:
        print(f"{count}\t{variant}\t{mem}\t{time}")
        mem = ""
        time = ""
```

himdel: Yeah, fair point, we should prefer tools in python over perl, I'll switch to this one :)

(+- the `in sys.stdin` bit .. perl `while (<>)` reads either from stdin, or from files passed in as args .. but python has `fileinput` for that)

MilanPospisil (approved):
LGTM :+1

PR #196, opened 2025-08-18
Fix help text

MilanPospisil:
[AAP-51735]

## Description

In the issue.

### DIFF
--- a/metrics_utility/management_utility.py
+++ b/metrics_utility/management_utility.py
@@ -62,7 +62,7 @@ class ManagementUtility(management.ManagementUtility):
         if commands_only:
             return commands
         else:
-            return f'Usage: manage.py <command> [options]\n{commands}'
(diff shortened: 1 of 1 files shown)


metrics_utility/management_utility.py
```diff
@@ -57,6 +57,13 @@ def execute(self):
         else:
             self.run_subcommand(subcommand, self.argv)
 
+    def main_help_text(self, commands_only=False):
+        commands = 'Commands: build_report, gather_automation_controller_billing_data'
+        if commands_only:
+            return commands
+        else:
+            return f'Usage: manage.py <command> [options]\n{commands}'
```
himdel: `manage.py` needs to respect the command name

`os.path.basename(sys.argv[0])` shoud work :)

himdel (approved):

PR #200, opened 2025-08-19
Only one message of generated report

MilanPospisil:
[AAP-51798]

Description in issue.

### DIFF
--- a/metrics_utility/base/collector.py
+++ b/metrics_utility/base/collector.py
@@ -172,15 +172,15 @@ class Collector:
         _now = now()
         original_since = since
         original_until = until
-        logger.info(f'Original since-until: {original_since} to {original_until}')
+        logger.warning(f'Original since-until: {original_since} to {original_until}')
 
         # Make sure that the endpoints are not in the future.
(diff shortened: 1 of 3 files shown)


metrics_utility/base/collector.py
```diff
@@ -172,15 +172,15 @@ def _calculate_collection_interval(self, since, until):
...
         original_since = since
         original_until = until
-        logger.warning(f'Original since-until: {original_since} to {original_until}')
+        logger.info(f'Original since-until: {original_since} to {original_until}')
 
         # Make sure that the endpoints are not in the future.
         if until is not None and until > _now:
             until = _now
-            logger.warning(f'End of the collection interval is in the future, setting to {_now}.')
+            logger.info(f'End of the collection interval is in the future, setting to {_now}.')
```
himdel: shouldn't most of these stay as warnings?

"hey we ignored your input and changed it" sounds like a warning to me

("original since-until" and "final since-until" and "progress info" do sound like info or debug, that change makes sense)

himdel (approved):

PR #204, opened 2025-08-28
generator: allow outputing main_host once per interval or weekly

himdel:
During build report perf tests, 1M hosts are generated, but the same set gets saved on each day - so extractors have to parse 31M host entries, not 1M, skewing the results.

=> Allow setting the frequency at which we output main_host tarballs from the generator.

This prevents having to parse 31M host entries when generating 1M, by only outputting the main_host data once per month in the generator.

```
MAIN_HOST_FREQUENCY=1 # default, output every day; used for 10k
MAIN_HOST_FREQUENCY=7 # output every 7 days; used for 100k
MAIN_HOST_FREQUENCY=0 # output once, at the end of the interval; used for 1M
```

(This only affects the generator, we should try to have the main_host collector only pick up hosts modified since, or something like that, for when something like this happens in the real world, but that's for #209.)

### DIFF
--- a/tools/perf/generator.py
+++ b/tools/perf/generator.py
@@ -400,14 +400,16 @@ Environment vars:
             return
 
         if table == 'main_host':
-            # main_host - only generate csvs once, not filtered by since/until; use for each daily tarball
+            # main_host - only generate csvs once, not filtered by since/until; output once, at end of the month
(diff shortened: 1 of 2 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #209, opened 2025-09-02
main_host collector: filter hosts by since & until (created OR modified)

himdel:
The idea is to reduce the amount of hosts we collect from *all the hosts*, ever,
to only hosts active during the collection period, but also to allow collecting inactive hosts too.

We still need inactive host data, just not necessarily daily, so, this becomes 2 different collectors:
* `main_host_daily` (new, optional, daily_slicing) 
* `main_host` (original, optional, limit_slicing)

The suggested way to run this is a daily cronjob with since/until enabled:

```
METRICS_UTILITY_OPTIONAL_COLLECTORS=main_host_daily
```

and then a separate weekly cronjob for just the all-hosts snaphot.

```
METRICS_UTILITY_OPTIONAL_COLLECTORS=main_host
METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR=true
```

Later PRs:

* update generator to allow both snapshots and dailies for main_host ... but also the other collectors

### DIFF
--- a/metrics_utility/automation_controller_billing/collectors.py
+++ b/metrics_utility/automation_controller_billing/collectors.py
@@ -23,6 +23,7 @@ from metrics_utility.base import register
 from metrics_utility.base.utils import get_max_gather_period_days, get_optional_collectors
 from metrics_utility.exceptions import MetricsException, MissingRequiredEnvVar
 from metrics_utility.library import CsvFileSplitter
+from metrics_utility.library.collectors.util import date_where
 from metrics_utility.logger import logger, logger_info_level
 
 from .prometheus_client import PrometheusClient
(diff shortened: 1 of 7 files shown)


metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -540,6 +539,36 @@ def main_host_table(since, full_path, until, **kwargs):
...
+        full_path,
+        until,
+        f"""
+        enabled='t'
+        AND (
+            (main_host.created >= '{since.isoformat()}' AND main_host.created < '{until.isoformat()}')
+            OR
+            (main_host.modified >= '{since.isoformat()}' AND main_host.modified < '{until.isoformat()}')
+            OR
+            (main_unifiedjob.created >= '{since.isoformat()}' AND main_unifiedjob.created < '{until.isoformat()}')
```
MilanPospisil: Should we add another OR with main_unifiedjob.finished so we got also all hosts that were used in jobs? Because we are collecting jobs data at the finish of the job, not at the job created.

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -541,6 +538,37 @@ def main_host_table(since, full_path, until, **kwargs):
...
+        return None
+
+    where = ' AND '.join(
+        [
+            "enabled='t'",
+            ' OR '.join(
+                [
+                    ' AND '.join([f"main_host.created >= '{since.isoformat()}'", f"main_host.created < '{until.isoformat()}'"]),
+                    ' AND '.join([f"main_host.modified >= '{since.isoformat()}'", f"main_host.modified < '{until.isoformat()}'"]),
+                    ' AND '.join([f"main_unifiedjob.created >= '{since.isoformat()}'", f"main_unifiedjob.created < '{until.isoformat()}'"]),
```
MilanPospisil: Maybe its slightly better way to filter it from started to finished? started is the real time when job started executing, created may be a bit misleading, but still very close to our result :)

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -531,7 +528,7 @@ def main_host_table(since, full_path, until, **kwargs):
                 ON main_organization.id = main_inventory.organization_id
             LEFT JOIN main_unifiedjob
                 ON main_unifiedjob.id = main_host.last_job_id
```
MilanPospisil: Im afraid this may be misleading and wrong?

ON main_unifiedjob.id = main_host.last_job_id

Imagine situation - host was included in jobs the day that satisfies since-until, BUT... the collection happens next day, for example in 1 in the morning and other job started just right after midnight, that will modified last_job_id to this particular job that is not satisfying since-until anymore and host will not appear in the search... What do you think?

We probably need to find different way how to connect hosts to running jobs.

One can be using jobhostsummary (which is large though :-( ) that connects host to job and is not being overwritten like last_job_id, so we can safely join there.

Another can be join running jobs with inventories they are using and lists of hosts from those inventories.

MilanPospisil: I think job host summary is the best way, joining it through inventory would work too, but it will be doing overcounting - not all hosts in inventory have to be used.

Job host summary has real usage of hosts. The info is also in events, but thats overkill :-)

metrics_utility/automation_controller_billing/collectors.py
himdel: TODO: remove main_unifiedjob condition, and remove until condition

MilanPospisil (approved):
Thanks for the completing feedback :-)

LGTM :+1
(Provided that tests will pass)

PR #210, opened 2025-09-02
Release 0.6.0

himdel:
Issue: AAP-51812

Release 0.6.0

PR #211, opened 2025-09-03
Bump versions to 0.6.1dev

himdel:
having tagged the 0.6.0 release (#210)
it's time to update the version to distinguish devel from 0.6.0.

---

(Technically, #188 obviates the need to change the version number everywhere, but .. one step at a time :).)


ShaiahWren (approved):
LGTM :) Unless the version needs to be changed anywhere else?

PR #213, opened 2025-09-16
Delete events in testathon script

MilanPospisil:
Very small fix, testathon script did not had events deletion at the script begining - simply because we did statistics on distinct usage of collections, roles, modules, the error was not discovered.


AlexSCorey (approved):
Looks good to me.

PR #215, opened 2025-09-18
Milan anonymized aggregations

MilanPospisil:
[AAP-54268]

## Description

added several rollups functions that computes metrics for anonymized data from dataframes that would be later joined from source tarballs.

Rollups directly corresponds to different collectors - events, jobhostsummaries, execution environments and jobs.

Also, in collectors were added some new info that was needed.

## Testing

You can simply run: uv run pytest -s {optionaly with path to particular test file}


metrics_utility/anonymized_rollups/collections_types.py
```diff
@@ -0,0 +1,11 @@
+# mocked data for now
+# pair collection name and ansible source type - validated, community, rh-certified etc.
+
+collections_types = {
```
himdel, line 4: You should be able to `collections_types = json.loads(os.path.join(os.path.dirname(__file__), 'collections.json')` once #225 is merged.

The format is the same, just `None` would have to be handled differently, maybe only .map after filtering for nonempty?
Oh, except `s/rh-certified/certified/`, the current values are `community` (for galaxy.ansible.com published), `certified` (for console.redhat.com published) and `validated` (for console.redhat.com validated).

(Doesn't have to be this PR though :))

metrics_utility/anonymized_rollups/events_anonymized_rollups.py
```diff
@@ -0,0 +1,244 @@
...
+
+    m = re.match(collection_regexp(), x)
+
+    if m:
+        return f'{m.groups()[0]}.{m.groups()[1]}'
+    else:
+        return None
+
+
+class Event_Anonymized_Rollups:
```
himdel: remove underscores from all these new classnames? (classnames are CamelCase, filenames are snake_case)

and, probably, it should also be singular, not plural..

if it helps... :)

```
for n in 1 2 3; do perl -i -npe 's/\b(([A-Z][a-z]+)+)_/$1/g' metrics_utility/**/*.py; done
perl -i -npe 's/\b(([A-Z][a-z]+)+)s\b/$1/g' metrics_utility/**/*.py
rename 's/s.py/.py/' metrics_utility/anonymized_rollups/*
git add metrics_utility/anonymized_rollups
```

himdel:
Looks good, all new code, or updates to new code, lgtm :)

Do I understand right that the `prepare_data` method is just a filtering step on the raw data,
and then the individual `_aggregate` functions convert those into rollup dataframes,
and the `base` method just returns a bunch of individual rollups from the same raw data?

So you'd first gather all raw data, pass it to `Event_Anonymized_Rollups.base(dataframe)`, and then save each value of the returned dict as a `.parquet` file?

This is still missing any functionality to actually merge 2 rollups, right?

---

> dict

Or, ideally, 1 rollup class should only return 1 kind of rollup, and this could be split into multiple classes?
Or is it really better to group them together?

There also seems to be a mix of values returned - a dict of dataframes sounds reasonable (if not 1 df per class), but dataframe.to_dict is surprising, why that?

metrics_utility/anonymized_rollups/collections_types.py
MilanPospisil: Thanks :-)

MilanPospisil:
> Looks good, all new code, or updates to new code, lgtm :)
> 
> Do I understand right that the `prepare_data` method is just a filtering step on the raw data, and then the individual `_aggregate` functions convert those into rollup dataframes, and the `base` method just returns a bunch of individual rollups from the same raw data?
> 
> So you'd first gather all raw data, pass it to `Event_Anonymized_Rollups.base(dataframe)`, and then save each value of the returned dict as a `.parquet` file?
> 
> This is still missing any functionality to actually merge 2 rollups, right?
> 
> > dict
> 
> Or, ideally, 1 rollup class should only return 1 kind of rollup, and this could be split into multiple classes? Or is it really better to group them together?
> 
> There also seems to be a mix of values returned - a dict of dataframes sounds reasonable (if not 1 df per class), but dataframe.to_dict is surprising, why that?

> This is still missing any functionality to actually merge 2 rollups, right?
We are not going to merge those rollups, thats up to SaaS team. We are going to only merge csv from tarballs into one dataframe and compute daily rollup.

If the data will be stored in parquet, or json...Im not sure, this is up to discussion. Maybe we can store everything as tarball csv or parquet and only then transform it into json when we are sending the data? Part of the data will be json though, some data makes no sense in tabular format, if its scalar value or list of module names for example.

But this is only POC, it needs additional work. Next will be reading the tarballs into single dataframe and storing data into whatever we will agree for. Another task will be reading the stored rollup, anonymize and send to ingress in whiever format we agree.

> Do I understand right that the `prepare_data` method is just a filtering step on the raw data, and then the individual `_aggregate` functions convert those into rollup dataframes, and the `base` method just returns a bunch of individual rollups from the same raw data?

Prepare data operates on single rows (at least I think :-D ), so it does not matter if we will do it per tarball or everything together at once.

himdel (approved):

himdel:
Approved, but do check the sonarqube results, `\w` is a good idea,
but those float comparisons .. I wouldn't expect them to pass on macs :)

You seem to be using `pytest.approx` already in some places, but it's missing from `metrics_utility/test/test_anonymized_rollups/test_events_modules_anonymized_rollups.py`

(And it's complaining about it *anyway* in `test_jobs_anonymized_rollups` .. but you can fix that by wrapping the constant instead of the variable .. it warns about `assert pytest.approx(rec_t1['job_waiting_time_average_in_seconds'], rel=1e-6) == 1.0` but should be ok with `assert rec_t1['job_waiting_time_average_in_seconds'] == pytest.approx(1.0)`.

PR #222, opened 2025-10-01
Update dependencies in pyproject.toml and uv.lock to latest versions:…

cshiels-ie:
… Django 4.2.21, psycopg 3.2.6, and setuptools 80.9.0. Increment revision in uv.lock to 3.


himdel (approved):

himdel:
Looks good, WDYT about enabling dependabot for these .. we didn't do it back in #98, but now might be the time..

PR #227, opened 2025-10-03
Extractors, Dataframes - allow filtering tarballs and extracted files by collection name

himdel:
Issue: AAP-52666

ExtractorDirectory & ExtractorS3 now take 2 extra arguments to `iter_batches`:
* `collections=`
* `optional=`

both can be `None`, in which case, no change, all files are processed as before and only filtered by enabled sheets.

When `collections` is set, we filter the set of tarballs to extract so that:
* tarballs created by 0.6.0 and earlier all get opened as before
* tarballs marked "-unknown" or "-config" are edge cases from #226 and we open them too
* tarballs ending in `-collection_key.tar.gz` get opened only if `collections` contains that collection
* (anything unrecognized, such as `source_tarball` when downloading from S3 will still get opened too)

When `collections` or `optional` are set, both are put together and passed as `enabled_set=` to `process_tarballs`.

Changed `process_tarballs` to take into account both enabled sheets and what the dataframe requests - so even if we open a tarball with only the wrong collections, we don't parse them.
And make `optional_sheets` optional there, as rollups won't have optional sheets but want all the files the dataframe asks for.

Changed `_safe_extract` to take also use `enabled_sheets` when provided and not even extract the files when not needed (and debug log the filenames too).

And unified the `indirect_nodes` vs `main_indirectmanagednodeaudit` naming, so that the name of the dataframe, the suffix of the tarball, and the collector key always match.

This might give up to 4x speedup for parsing files when multiple dataframes are enabled, by not unpacking the tarballs at all if we don't have to, and not unpacking & parsing all the files inside again and again.

(This helps build_report speeds, but also prepares ground for rollups using the same mechanism - that part will come in a third PR.)

### DIFF
--- a/metrics_utility/automation_controller_billing/dataframe_engine/dataframe_collection_status.py
+++ b/metrics_utility/automation_controller_billing/dataframe_engine/dataframe_collection_status.py
@@ -10,7 +10,8 @@ class DataframeCollectionStatus(Base):
         dataframe = None
 
         for date in self.dates():
-            for data in self.extractor.iter_batches(date=date):
+            # collections=None because data_collection_status is in every tarball, and none of them are named after it
+            for data in self.extractor.iter_batches(date=date, collections=None, optional=['data_collection_status']):
                 batch = data['data_collection_status']
                 if batch.empty:
                     continue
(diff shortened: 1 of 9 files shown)


cshiels-ie (approved):

himdel:
Added a tweak to not ignore files which don't follow any of the naming patterns,
for compatiblity and so that `source_tarball` downloaded from S3 keeps working.

(Will add logic for S3 to not download these when not matching in a separate PR.)

PR #233, opened 2025-10-07
old replacement with new collector

MilanPospisil:
[AAP-54269]

## Description

Collectors were replaced by new ones, but old ones kept as deprecated. After perf tests, they can be removed completely.

## Testing

uv run pytest -s

More in depth test:
run awx in your local machine

in your standalone metrics-utility, run:
cd testathon
python3 testathon_data_prepare.py

Data will be pushed into awx DB.

Now attach to awx container and run gather command (with span over todays date), then build command from it. You should see the report.

# Cursor analysis

# Replace Deprecated Collectors with New Versions

## 🔄 What Changed
Replaced deprecated collectors:

- `main_jobevent` → replaces **main_jobevent_deprecated**  
- `job_host_summary` → replaces **job_host_summary_deprecated**

---

### **main_jobevent**

**Removed:**
- `main_jobhostsummary_id`
- `main_jobhostsummary_created`

**Added:**
- `job_finished`
- `uuid`
- `parent_uuid`
- `start`
- `end`
- `task_uuid`
- `warnings`
- `deprecations`
- `playbook_on_stats`
- `job_failed`
- `job_started`

**Unchanged core fields:**
`id`, `created`, `modified`, `job_created`, `event`, `task_action`,  
`resolved_action`, `resolved_role`, `duration`, `failed`, `changed`,  
`playbook`, `play`, `task`, `role`, `job_remote_id`, `host_remote_id`, `host_name`

---

### **job_host_summary**

**Columns:**  
Same effective set as before:
- IDs and timestamps
- `host_name`
- Counters (`dark`, `failures`, `ok`, `skipped`, `ignored`, `rescued`)
- `job_*` fields
- `organization`, `inventory`, `project`
- `ansible_host_variable`, `ansible_connection_variable`

**Selection strategy change:**  
Now filters by jobs that **finished in the window** (`main_unifiedjob.finished`)  
instead of using summary **modified timestamps**.

---

### **main_jobevent → DataframeContentUsage**

**Uses:**
- `host_name`
- `task_action` / `resolved_action` / `resolved_role`
- `duration`
- `role`
- `job_remote_id`

✅ **No breakage**  
Removed fields were unused; added fields are ignored.  
Grouping and renaming remain unchanged.

---

### **job_host_summary → DataframeJobhostSummaryUsage**

**Uses:**
- `host_name` (normalized via `ansible_host_variable`)
- Counters: `dark`, `failures`, `ok`, `skipped`, `ignored`, `rescued`
- `created`
- `organization_name`
- `job_template_name`
- `job_remote_id`
- Optionally: `job_created`

✅ **No breakage**  
All required columns exist; extra fields are ignored.  
Aggregation logic remains the same.

---

### **Time-Window Semantics**
- `main_jobevent`: Events tied to **jobs finished** within the window.  
- `job_host_summary`: Driven by **finished window** instead of **modified**.

📌 *Possible slight shifts in row inclusion at boundaries.*  
Does **not** affect dataframe logic.

---

## ⚡ Performance
Both collectors now **filter by finished jobs first**,  
reducing scanned rows on large instances.

---

## 🧱 Backward Compatibility
- **Build report flows (CCSP, CCSPv2):** Continue to work.  
- Added columns → harmless.  
- Removed columns → weren’t used.

⚠️ *External tools depending on*  
`main_jobevent.main_jobhostsummary_id` or  
`main_jobevent.main_jobhostsummary_created`  
*will need updates.*

---

## 🧪 Test Coverage
Gather tests now:
- Assert **exact headers** of `main_jobevent.csv` and `job_host_summary.csv`
- Aligned with the **new schemas**
- Ensures full compatibility with the **build process**

# Performance Analysis for Job Event

## ✅ New Version (`main_jobevent`)
**Pros**
- Uses constant `(job_id, job_created)` pairs → strong **partition pruning**.
- **Removes expensive join** to `main_jobhostsummary`, significantly reducing I/O.
- Limits scope to **finished jobs** within the time window.
- Generally **faster** for typical daily/hourly slices (hundreds of jobs).
- Simpler and more predictable query plan.

**Cons**
- If the time window contains **thousands of jobs**, the large
  `IN (VALUES …)` clause increases **planning time** and query text size.
- Currently missing the `event IN (...)` filter used in the old version,
  which could reduce the number of scanned rows.

---

## ⚠️ Deprecated Version (`main_jobevent_deprecated`)
**Pros**
- Already filters events (`runner_*`), limiting output size.
- Uses `main_jobhostsummary.modified` (which is indexed) for slicing.

**Cons**
- Joins to `main_jobhostsummary`, which can be a **large table**, causing:
  - Heavy **join cost** (I/O, memory, and CPU).
  - Broader scans due to extra table lookups.
- **Weaker partition pruning** because filtering happens through a join/CTE
  rather than constant predicates on `job_created`.
- Touches unnecessary data when only job event details are needed.

---

### DIFF
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -57,6 +57,7 @@ VALID_COLLECTORS = {
     'main_jobevent_service',
     'execution_environments',
     'job_host_summary_deprecated',
+    'main_jobevent_deprecated',
 }
 
 VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -56,7 +56,7 @@ VALID_COLLECTORS = {
     'unified_jobs',
     'main_jobevent_service',
     'execution_environments',
-    '',
+    'job_host_summary_deprecated',
 }
 
 VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -58,6 +58,7 @@ VALID_COLLECTORS = {
     'execution_environments',
     'job_host_summary_deprecated',
     'main_jobevent_deprecated',
+    '',
 }
 
 VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
--- a/metrics_utility/management/validation.py
+++ b/metrics_utility/management/validation.py
@@ -54,7 +54,6 @@ VALID_COLLECTORS = {
     'main_indirectmanagednodeaudit',
     'total_workers_vcpu',
     'unified_jobs',
-    'job_host_summary_service',
     'main_jobevent_service',
     'execution_environments',
     '',
--- a/metrics_utility/automation_controller_billing/collectors.py
+++ b/metrics_utility/automation_controller_billing/collectors.py
@@ -214,122 +214,6 @@ def yaml_and_json_parsing_functions():
     return query
 
 
-@register('job_host_summary', '1.2', format='csv', description=_('Data for billing'), fnc_slicing=daily_slicing)
-def job_host_summary_table(since, full_path, until, **kwargs):
-    disable_job_host_summary_str = os.getenv('METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR', 'false')
-    disable_job_host_summary = False
-    if disable_job_host_summary_str and (disable_job_host_summary_str.lower() == 'true'):
-        disable_job_host_summary = True
-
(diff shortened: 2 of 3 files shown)


metrics_utility/management/validation.py
```diff
@@ -54,10 +54,10 @@
     'main_indirectmanagednodeaudit',
     'total_workers_vcpu',
     'unified_jobs',
-    'job_host_summary_service',
     'main_jobevent_service',
     'execution_environments',
-    '',
```
himdel, line 60: The empty string is there because otherwise validation rejects empty optional collectors.

(So, we need to keep it, or fix validation not to need it.)

MilanPospisil: ah, I deleted it by accident. Nice catch!

PR #236, opened 2025-10-09
Properly clean the tests files

MilanPospisil:
Tests were not properly cleaning the data and caused unstaged git changes after tests runs, now they clean everything.


himdel (approved):
Good catch, everything with the 00000* uuid is safe to remove :+1:

PR #245, opened 2025-10-27
Add conf_setting data to db mocks

himdel:
Adding a `tools/docker/conf_setting.sql`, with:
* `LICENSE` - matching the mock `mock_awx/awx/conf/license/__init__.py` license, which was itself based on https://github.com/ansible/awx/blob/devel/awx/main/utils/licensing.py#L88
* `INSTAL_UUID` & `TOWER_URL_BASE` - matching the mock `mock_awx/settings/__init__.py` settings.

(And wiring into `make compose` & the github CI.)

This should help with testing #242 which switches from asking awx to looking for these in the db.

---

And update `run-*` scripts to no longer try to create their own db environment, instead using the `make compose` one.
(They no longer matched because only compose is getting the table data; `run-s3-gather-build` already went that way.)

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: "Import SQL dump"
         run: |
-          cat tools/docker/{roles,latest,main_hostmetric,main_jobhostsummary}.sql | sudo su - postgres -c psql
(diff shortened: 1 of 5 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #248, opened 2025-10-29
AAP-54476 : library.collectors - move collectors to library

himdel:
Issue: AAP-54476

Move all collectors into `metrics_utility.library.collectors`. (Technically, copy, removing the originals is a second PR.)

A collector is a function decorated with `@collector`, it either returns a dict, or a list of csv filenames (from `copy_table`),
it always takes a `db=` parameter if it's supposed to talk to a db, or credentials for other services,
it never reads environment variables directly, always using params,
it takes `since=` and `until=` timestamp-with-timezone params if the collector supports collecting within time contraints, `since` is included (`>=`), `until` excluded (`<`),
it only takes args it uses, no extra `**kwargs`.

The `@collector` decorator wraps these into a class instance such that `collector_fn(params).gather()` actually calls the original `collector_fn` with `params`. This is to allow splitting collector config (which should happen outside of global tempdir handling) from actual gather (which should happen while holding a db lock).

Notes:

* `output_dir=` is needed for compatibility with the cli, but omitting it will work too (and create a new tempdir per collect)
* `output_file=` (unused but supported in `copy_table`) allows streaming the data without creating tempfiles - just writes to a filehandle object
* no cleanup happens for the collected files (except for the cleanup done by cli) - this will be the role of the library tempfile helper (wip, separate issue)

Controller collectors (`metrics_utility.library.collectors.controller.*`)

* `config(db, billing_provider_params).gather() -> Dict`
* `execution_environments(db, [output_dir]).gather() -> [filenames]`
* `job_host_summary(db, since, until, [output_dir]).gather() -> [filenames]`
* `job_host_summary_service(db, since, until, [output_dir]).gather() -> [filenames]`
* `main_host(db, [output_dir]).gather() -> [filenames]`
* `main_indirectmanagednodeaudit(db, since, until, [output_dir]).gather() -> [filenames]`
* `main_jobevent(db, since, until, [output_dir]).gather() -> [filenames]`
* `main_jobevent_service(db, since, until, [output_dir]).gather() -> [filenames]`
* `unified_jobs(db, since, until, [output_dir]).gather() -> [filenames]`

Other collectors (`metrics_utility.library.collectors.others.*`)

* `total_workers_vcpu(cluster_name, metering_enabled, prometheus_url, ca_cert_path, token) -> Dict`

---

Separate PR (#260):

The original collectors file retains all the collector functions, original decorators, any get_optional_collectors logic,
but otherwise the functions end up calling the library functions.

Later:
* move from f-strings to named query params - copy_table can handle it now; drop psycopg2, use safe sql
* why the extra backslash replace on event_data?


cshiels-ie (approved):
LGTM 👍

PR #249, opened 2025-10-31
Added file splitting due to size limits with Segment.com 

cshiels-ie:
Enhance StorageSegment to support chunked data transmission for Segment analytics. Implement size limits for regular and bulk messages, and add methods to calculate data size and split data into manageable chunks. Update the put method to handle large datasets by sending them in chunks while maintaining debug logging for better traceability.

### DIFF
--- a/metrics_utility/library/storage/segment.py
+++ b/metrics_utility/library/storage/segment.py
@@ -1,35 +1,213 @@
 import datetime
+import json
+import sys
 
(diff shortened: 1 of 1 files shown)


himdel (approved):
Looks like it should work, but I wonder if there's places where splitting in the middle doesn't make sense. But that will depend on the data...

Approved :)

PR #250, opened 2025-10-31
AAP-57241: Milan anonymized hardening

MilanPospisil:
[AAP-[57241](https://issues.redhat.com/browse/AAP-57241)]

## Description

This PR changes many things:

1) Jobs that never started are counted and not filtered (for jobs and events rollups)

2) list of modules uset to automate are no longer in statistics (can be computed from overall statistics of modules) - freeing up the space in json

3) hash is shortened in half

4) Events and job host summaries are now done in batches - in prepare function, some pre agggregates are computed. For events, we are looking for occurences of specific events, thus reducing possible reruns into 1 row (except reruns that are scattered through multiple batches, but that is rare).  So the same must be done again on overall data (to group rows that belong to the same task_uuid on specific host and aggregate seen events again.

5) Job host summaries prepare function simply sums up task counts by template in batches - again reduce memory requirements

6) The events data strings are categorized - this does the pandas easily - it can reduce the memory requirements - but this is done after all batches are concatenated (categorization does not work accross different dataframes that are then concatenated). For even more performant categorization (accross batches), we can do some our own implemenation in the future. But since the whole data needs some memory and operations over data too, its good to reduce them as much as possible.

7) Sanitize json - all NaN values are replaced with null (NaN are not valid for json)

8) metrics_utility/test/test_anonymized_rollups/test_multiple_files.py
This is testing reading from multiple dataframes (testing prepare and concatenation), also it produces the final json and asserts.

9) total number of jobs executed is added.

10) Output json is rearranged, so there are not nested objects other than key : object or key : list of objects

Any other required tests that reflect changes are updated.

## Testing

Most important test:
uv run pytest -s metrics_utility/test/test_anonymized_rollups/test_multiple_files.py

This test will create in out/rollups/year/month/day/anonymized.json

The json that can be sent to the ingress.

### DIFF
--- a/metrics_utility/anonymized_rollups/anonymized_rollups.py
+++ b/metrics_utility/anonymized_rollups/anonymized_rollups.py
@@ -9,6 +9,7 @@ from pandas import DataFrame
 from metrics_utility.anonymized_rollups.base_anonymized_rollup import BaseAnonymizedRollup
 from metrics_utility.anonymized_rollups.events_modules_anonymized_rollup import EventModulesAnonymizedRollup
 from metrics_utility.anonymized_rollups.execution_environments_anonymized_rollup import ExecutionEnvironmentsAnonymizedRollup
+from metrics_utility.anonymized_rollups.helpers import sanitize_json
 from metrics_utility.anonymized_rollups.jobhostsummary_anonymized_rollup import JobHostSummaryAnonymizedRollup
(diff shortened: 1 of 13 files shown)


ShaiahWren:
@MilanPospisil the changes are very robustly tested, for sure. It seems like the only extra confidence test would be to do a comparison between devel and this pr branch with real data.

ShaiahWren (approved):

himdel (approved):

PR #257, opened 2025-11-04
config.json - take version from main_instance only

himdel:
Follows #242 

Change `get_controller_version_from_db` to only use the `main_instance` table, or null.

Fixups:

* `get_last_entries_from_db` returns a Dict, adjust type info
* remove mentions of a `VERSION` `conf_setting` entry; and `sku` is part of license info

### DIFF
--- a/metrics_utility/automation_controller_billing/helpers.py
+++ b/metrics_utility/automation_controller_billing/helpers.py
@@ -1,7 +1,7 @@
 import json
 
 from itertools import chain
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Tuple
 
 import pandas as pd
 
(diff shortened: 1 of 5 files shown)


metrics_utility/automation_controller_billing/helpers.py
```diff
@@ -64,48 +64,29 @@ def get_config_and_settings_from_db() -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
-    except Exception as e:
-        logger.error(f'Error getting AWX/Controller version from database: {e}')
-        logger.error('Returning "Database error" as fallback for controller version')
-        return 'Database error'
-    return 'No data found'
+
+def get_controller_version_from_db() -> str:
+    """Get AWX/Controller version from the main_instance DB table."""
+    return _fetch_one(
+        connection,
```
AlexSCorey, line 79: I see you removed the try except blocks.  Shouldn't we catch errors if the db isn't reachable?

AlexSCorey (approved):
@himdel addressed all my concerns offline.

PR #259, opened 2025-11-05
[AAP- 54910] Removes references to awx for db locks

AlexSCorey:
[AAP-54910]

## Description

This PR is part of the metrics library epic.  It replace database locking functionality which was provided by Django Ansible Base with our own 
## Testing

To test you could kick off 2 gather jobs in quick succession.  The first 1 should pass and the second should fail.

---

> [!NOTE]
> Introduce an internal Postgres advisory lock helper and switch collectors to use it, removing AWX-specific logic and adding unit tests.
> 
> - **Library**:
>   - Add `metrics_utility.library.lock.lock` implementing Postgres advisory locks using `hashtext` and `pg_*` functions.
>   - Export `lock` in `metrics_utility.library.__init__`; remove the previous stub lock from `utils`.
> - **Collectors**:
>   - `metrics_utility/base/collector.py`: use `lock('gather_analytics_lock', wait=False, db=connection)` during gather.
>   - `metrics_utility/automation_controller_billing/collector.py`: use `lock(...)` for billing gather and finalize; remove AWX-specific advisory lock imports and helper.
> - **Tests**:
>   - Add `metrics_utility/test/library/test_lock.py` to validate string-key hashing and lock acquisition/contension behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9bd4dd6ca92820a11449b5bf0c2922359c053a3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


AlexSCorey:
cursor review

AlexSCorey:
cursor review

MilanPospisil:
Great job Alex!

However, this seems to be a lot of functionality that we do not actually need. We should wait for @himdel when he returns from sickness leave, if I do understood it correctly or Im just wrong here and I did not understood the Martins intentions.

Because from first look, this seems to have support for SQL Lite and for Django, while metrics-utility is not using neither.

So we can simply call only basic functions of advisary lock from postgres. But I will look at it tomorrow in more depth :)

No worries Alex, we will clarify this :)

metrics_utility/automation_controller_billing/collector.py
```diff
@@ -97,9 +89,13 @@ def registered_collectors(cls, module=None):
         return base.Collector.registered_collectors(collectors)
 
     @contextlib.contextmanager
-    def _pg_advisory_lock(self, key, wait=False):
+    def _pg_advisory_lock(
+        self,
+        *args,
+        **kwargs,
+    ):
```
himdel:
```suggestion
    def _pg_advisory_lock(self, key, wait=False):
```

metrics_utility/automation_controller_billing/collector.py
```diff
@@ -97,9 +89,13 @@ def registered_collectors(cls, module=None):
...
         """Use awx specific implementation to pass tests with sqlite3"""
-        with advisory_lock(key, wait=wait) as lock:
+        with advisory_lock(*args, **kwargs) as lock:
```
himdel:
```suggestion
        with advisory_lock(key, wait) as lock:
```

metrics_utility/base/collector.py
```diff
@@ -309,28 +310,22 @@ def _add_collection_to_package(self, collection):
...
-                if wait:
-                    cursor.execute('SELECT pg_advisory_lock(%s);', (resource_key,))
-                else:
-                    cursor.execute('SELECT pg_try_advisory_lock(%s);', (resource_key,))
-                acquired = cursor.fetchall()[0][0]
-                yield acquired
-            finally:
-                cursor.execute('SELECT pg_advisory_unlock(%s);', (resource_key,))
-                cursor.close()
+    def _pg_advisory_lock(self, key, shared=False, wait=False, db_alias=DEFAULT_DB_ALIAS):
```
himdel: shared & db_alias are not needed here

MilanPospisil: Shared will be always false for our utility, because we want exclusive access - only one utility can read the db at time because of performance.

Shared True is for different purposes, typicaly reader/writer sync. In this scenario, readers can read in parallel (and this is not we want in utility because of performance). But writers can access only one at time. When the writer has lock, no reader can read. When the readers have the lock, writer can not write.

So Writers are acquiring lock with Shared=False. Readers are acquiring as Shared=True.

But in our situation, we want exclusive access even for readers, thus Shared=False every time.

MilanPospisil: DB alias can be deleted, default will be used.

metrics_utility/base/db_lock_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+
+def _validate_tuple(key):
+    if len(key) != 2:
+        raise ValueError('Tuples and lists as lock IDs must have exactly two entries.')
+    if not isinstance(key[0], int) or not isinstance(key[1], int):
+        raise ValueError('Both members of a tuple/list lock ID must be integers')
+    return True
+
+
+def _validate_and_create_generated_key(key):
```
MilanPospisil: This is ok, but we will be probably using strings only anyway.

metrics_utility/base/db_lock_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+from zlib import crc32
+
+from django.db import DEFAULT_DB_ALIAS, connection, connections
+
+from metrics_utility.logger import logger
+
+
+@contextmanager
+def _pglocks_advisory_lock(key, shared=False, wait=False, db_alias=DEFAULT_DB_ALIAS):
+    function_name, release_function_name = _generate_function_name(wait, shared)
```
MilanPospisil: This can be simplified.

1) We do not need shared, shared will be always false, thus less variability for function name.

I would rather use direct function names than having function_name and release_function_name as variables, this is less readeble and with lower number of variability, not really needed.

Key can be always treated as string (it is enough for our purposes), so you can simplify also this and remove:

_generate_function_name
_validate_tuple_
_validate_and_create_generated_key

Only string to integer (crc32) is needed.

metrics_utility/test/test_db_lock_helpers.py
```diff
@@ -0,0 +1,299 @@
+from concurrent.futures import ThreadPoolExecutor
```
MilanPospisil: After short discussion with Martin.

These tests looks tests mocks, not real locking.

What about one simple single test that:

with advisory_lock('abc', wait=False):
  with pytest.expect(Exception) as e:
    with advisory_lock('abc', wait=False):
      pass
      
 This will test locking with Wait=False. Second lock grab should fail. We can not simply test Wait=True without parallelism, but that is outside of this PR.

AlexSCorey: @MilanPospisil so these are unit tests and do have an actual db connection.  This test does the same thing, but it does mock the db `test_lock_convention_integration`

AlexSCorey:
cursor review

metrics_utility/base/db_lock_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@contextmanager
+def advisory_lock(
+    key,
+    wait,
+):
+    # Acquire the lock and yield to caller
+    with _pglocks_advisory_lock(key, wait) as internal_lock:
+        yield internal_lock
```
himdel:
```suggestion
```

this just calls the method above, with the same params, and does what it already does by itself, remove this and rename the above method to `advisory_lock`?

metrics_utility/automation_controller_billing/collector.py
```diff
@@ -99,7 +91,7 @@ def registered_collectors(cls, module=None):
     @contextlib.contextmanager
     def _pg_advisory_lock(self, key, wait=False):
         """Use awx specific implementation to pass tests with sqlite3"""
-        with advisory_lock(key, wait=wait) as lock:
+        with advisory_lock(key, wait) as lock:
             yield lock
```
himdel: Neither of these `_pg_advisory_lock` *methods* are needed anymore, just change any code calling them to call `advisory_lock` directly

metrics_utility/base/collector.py
```diff
@@ -310,27 +310,21 @@ def _add_collection_to_package(self, collection):
 
     @contextlib.contextmanager
     def _pg_advisory_lock(self, key, wait=False):
```
himdel: > Neither of these _pg_advisory_lock methods are needed anymore, just change any code calling them to call advisory_lock directly

(and here)

metrics_utility/test/test_db_lock_helpers.py
```diff
@@ -0,0 +1,115 @@
...
+            with advisory_lock(f'{key}', False) as acquired:
+                if acquired:
+                    lock = f'Thread {thread_id} acquired lock. Key: {key}'
+                    return lock
+                else:
+                    lock = f'Thread {thread_id} did not, and should not, acquire lock. Key: {key}'
+                    return lock
+        except Exception as e:
+            lock = f'Thread {thread_id} encountered error: {e}. Key: {key}'
+            return lock
```
himdel: please don't use threads if we don't have to

this could be as simple as 

```suggestion
    def test_acquire_lock(self):
        with advisory_lock('test', False) as acquired:
            assert acquired is not None
            with pytest.raises(Exception) as excinfo:
                with advisory_lock('test', False) as acquired2:
                    assert False, "this should be unreachable"
```

feel free to remove `test_collector_lock_behavior` & `test_gahter_calls_lock` - in the library, collectors don't take locks, the task does

does `test_string_key_conversion` do anything anymore? since this is always a string?

the rest, up to you, but no `ThreadPoolExecutor` please :)

himdel:
OK, this is good, couple of comments above, but this is mergeable :tada: :)

Are you intentionally leaving the AAP-54910 part to a second PR?

If not - the `advisory_lock` helper needs to take an extra `db` param,
must NOT touch `connections['default']` but use that `db` instead.

And move from helpers to `metrics_utility/library/lock.py`, and rename to just `lock(db=..., key=..., wait=False)` .

(And callers need to pass in `connection` for the db param (`from django.db import connection` ... `library.lock(db=connection, ......)`))

metrics_utility/test/test_db_lock_helpers.py
MilanPospisil: One thing that needs to be considered is also what is the scope of the lock - I believe its database session. This means, if there will be one DB session, second inner lock call will actually pass (this is called reentrant locking, its common and chatgpt tells me that this is the case of postgres lock, we may to double check this).

If this is true, the suggestion from Martin will not throw exception and will not work as intended, in this case:

We must make sure that we are using different DB sessions for each lock.

Simple theory behind:
Every lock has owner, if the owner has lock and wants it to acquire the lock again, it will succeed. There is lock counter that will count the count of acquire and every unlock will lower this count. So:

lock
      lock (will pass for same session, now locked for second time)
      unlock (we are still locked!!)
unlock (now we are finally unlocked)

So Im not sure if we will be able to do it without threads :-( Because if I understand it correctly, DB connections are reused for thread - so each thread holds single session - thus if you will use lock twice inside one thread (DB session), it will acquire lock without problems. But if there is simple way how to use two sessions and pass them to acquire lock, this would work...

If not, I suggest leave it be and approve it (unless we come up with simple solution), locking is not crucial for our app anyway.

himdel: Nah, no worries, it still works :) 

The db requests are all serialized through one thread anyway.

Tested with...

```python
import pytest

from django.db import connection

from metrics_utility.library import lock


def test_acquire_lock():
    with lock(connection, 'abc', False) as acquired:
        assert acquired is not None
        with pytest.raises(Exception):
            with lock(connection, 'abc', False):
                assert False, 'unreachable'
```

himdel (approved):

PR #260, opened 2025-11-06
cli - use library.collectors

himdel:
Issue: AAP-54476 / AAP-57129 / AAP-66077

This is the second part of #248: this makes sure we use the collectors moved to the library from the cli.

It also adds `until_slicing` - which will use the last date of `daily_slicing`, unlike `limit_slicing` which uses today - and uses it for execution environments.

Also remove dead full_sync logic,
stop passing unused max_gather_size,
use init_tmp_dir helper from the library,
and adjusts test imports.

EDIT: most of this was merged in #339, what remains may or may not be useful, draft until I can revisit this


SherinV:
This is to be closed once [#339](https://github.com/ansible/metrics-utility/pull/339) goes in.

himdel:
Split and close...

### Commits on Nov 7, 2025

* logger - rename debug() to logger_debug()
* remove unused hook & settings
* collector.gather - remove unused dest, call init_tmp_dir from library
* cosmetic - inline get_last_entries call
* collection.gather - remove unused max_data_size
* Move collectors
* slicers - only require key
* collectors - pass in output_dir instead of full_path, pass in billing_provider_params for config, remove collection_type, unused
* test changes
* pass billing params to base collector, to prevent key error in tests
* drop unused full_sync
* cosmetic - no need fo f-string
* until_slicing - fix to use the last day, not the first non-collecting day
* _gather_config - remove code patching config collector output with billing_provider_params
* fixup imports to use library.collectors.{controller,others}.*
* last _last_gathering

himdel:
Changes applied in the gather-only branch

himdel:
(already closed, mentioning #540

PR #268, opened 2025-11-07
Move testathon/ to tools/testathon/

himdel:
Move `testathon/` to `tools/testathon/`,
and update paths to match.


cshiels-ie (approved):
Makes sense 👍

PR #274, opened 2025-11-13
Change exposed port for metrics-utility

MilanPospisil:
[AAP-XXXXX]
Dont forget to link back issue to PR.

## Description

Exposed port should be changed so it does not collide with the same exposed port for metrics service DB, so we can run both on local machine and dont have to manualy start/stop based on what we are working at.

CI test fails because pytest.yml needs to set the default postgres port, but its pull request target, so the changes are valid after PR is merged. If there will be any problems after merge, we will revert this PR.

## Testing

export METRICS_UTILITY_SHIP_PATH="./out"
export METRICS_UTILITY_SHIP_TARGET="directory"

export METRICS_UTILITY_OPTIONAL_COLLECTORS="execution_environments,job_host_summary_service,main_host,main_indirectmanagednodeaudit,main_jobevent,main_jobevent_service,unified_jobs"

uv run ./manage.py gather_automation_controller_billing_data --ship --since=2025-06-13 --until=2025-06-14

Failure on collecting indirect data is expected, since we have older version of awx from winter before indirect node counting (and we are missing automatic tests here probably).

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -92,6 +92,7 @@ jobs:
           METRICS_UTILITY_BUCKET_REGION: "us-east-1"
           METRICS_UTILITY_BUCKET_SECRET_KEY: "myusersecretkey"
           METRICS_UTILITY_DB_HOST: "localhost"
+          METRICS_UTILITY_DB_PORT: "5432"
         run: |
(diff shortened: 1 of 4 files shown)


himdel (approved):
CI will work after merge

PR #281, opened 2025-11-21
Milan fix segment batches

MilanPospisil:
[AAP-59062]

Reimplementing the splitting algorithm so it matches anonymized data.

It is splitting data into chunks. Each chunk is send to segment.com

The algorithm is simple, deterministic and requires that our data is in shape below - which is not that hard to maintain.

The data are in shape as follows:

{
   'statistics' : {...dict...}
   'modules' : [...list...]
}
The main data structure must be dict.
It can have multiple of those key:values in main dict, but everything must be either dict or list.

The algorithm for splitting is follows:
1) if the size of the data is smaller than limit, return only one chunk with all the data
2) If it is larger, begin splitting. 
     a) every key:value in main dict is separate chunk
     b) If value is dict, the whole dict goes into chunk, no splitting here, we are using dicts only for some overall summaries and should not be large
     c) if value is list, begin splitting, lists can have thousands of items inside. List are split into chunks, so each chunk will be lower than limit, so the list will be send in multiple requests to segment.com.
     
 # Testing
 uv run pytest -s metrics_utility/test/library/test_storage_segment.py
 
 But real test will come in integration tests where we will be sending real data.


cshiels-ie (approved):
LGTM

PR #284, opened 2025-11-24
Optimize service event query

MilanPospisil:
[AAP-59250]

Speedup for main_jobevent_service and main_jobevent in library using LATERAL.

main_jobevent service now reduces number of pairs in jobs:

    where_clause = f'(e.job_id, e.job_created) IN (VALUES {pairs_sql})'
    
This will contain all jobs finished in since-until, which can be tens of thousands of jobs!
 
 This can easily bloat. We introduce new approach. Since we can not join on jobs according to job_created, because in order to work with partitions prunning on job_created (job_created is houlrly partition in event table), we need to do very clear literal explicit filter on event table (thus WHERE with contstant job_created).
 
 Improvements: 
 job_created is simplified - awx uses hourly partitions, so we split into hourly job_created and put it into set, this keeps max hundreds of hours
 
 Then we also makes groups of nearby hours to reduce it even more, for example 2025-06-01 10:00:00 and 2025-06-01 11:00:00 will be concatenated and the resulted range filter will be job_created => min AND job_created < max.
   
Also removed cyclic import.

### DIFF
--- a/metrics_utility/library/collectors/controller/main_jobevent_service.py
+++ b/metrics_utility/library/collectors/controller/main_jobevent_service.py
@@ -4,8 +4,6 @@ from ..util import collector, copy_table
 @collector
 def main_jobevent_service(*, db=None, since=None, until=None, output_dir=None):
     # Use the table alias 'e' here (you alias main_jobevent as e in the FROM)
-    # FIXME: suspicious replace, why not e.event_data::jsonb->>'foo'
-    event_data = r"replace(e.event_data, '\u', '\u005cu')::jsonb"
 
     jobs_query = """
         SELECT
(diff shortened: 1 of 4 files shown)


metrics_utility/library/collectors/controller/main_jobevent_service.py
```diff
@@ -26,12 +31,84 @@ def main_jobevent_service(*, db=None, since=None, until=None, output_dir=None):
...
+                job_id INTEGER PRIMARY KEY
+            )
+        """)
+
+        # Insert unique job_ids
+        job_ids_set = set(job_id for job_id, _ in jobs)
+
+        # Use execute_values for efficient bulk insert
+        try:
+            from psycopg2.extras import execute_values
```
himdel: This is problematic.

psycopg2 might be available even when connection uses psycopg3 - specifically in metrics service dispatcher installs v2 too

We should probably use the same way of determining v2 vs v3 as in copy_table .. `if hasattr(cursor, 'copy_expert') and callable(cursor.copy_expert):`

(importing psycopg2.extras is fine, but only if connection is v2 too)

MilanPospisil: Is the new change acceptable?

himdel (approved):

PR #237, opened 2025-10-09
library.dataframes: Move existing dataframe classes to library, split loading from logic

himdel:
Issue: AAP-54468

Moving dataframes to library, and updating so that:

* the superclass of all dataframes is `BaseDataframe`, which no longer has any of the logic around merging and aggregation of the more complex dataframe - that part was moved to `BaseTraditional`

A `BaseDataframe` instance has these properties and methods:

* `.rollup` - None, or a pandas dataframe with the acumulated rollup data (what `build_dataframe` returns)
* `.add_raw(df)` - add raw (csv) data - calls `.prepare()`, `.group()` and passes the result to `.add_rollup()`
* `.add_rollup(df)` - add a rollup to the existing one - calls `.merge()`
* `.merge(old, new)` - concatenates 2 non-empty dataframes; overriden in `BaseTraditional`
* `.prepare(df)` - apply any preprocessing on the raw data (noop by default)
* `.group(df)` - apply any grouping & aggregation to the data (noop by default)
* `.regroup(df)` - reapply grouping after it was already done (used by deduplication; noop by default)
* `.empty()` - return an empty dataframe; overriden in `BaseTraditional` to know which fields it has.
* `.postprocess(df)` - apply any postprocessing after all data is loaded (reset_index by default)
* `.from_tarballs(batches)` - old build_dataframe - iterates over batches of raw data, calls `.add_raw()`, `.postprocess()`
* `.to_csv(f)` - saves as csv to a filename or file handle
* `.to_json(f)` - saves as csv to a filename or file handle
* `.to_parquet(f)` - saves as csv to a filename or file handle

Also, 3 loader functions are exported - `from_csv(f)`, `from_json(f)` & `from_parquet(f)` - each loads a filename/filehandle and return a raw dataframe (to be used with `add_raw` or `add_rollup`).

`BaseDataframe` can also be used as a superclass for anonymized rollups, method signatures match, wiring it in is a separate PR.

A `BaseTraditional` dataframe overrides `empty` and `merge` to implement the behaviour implemented by the 3 complex ccsp rollups, implements `cast_dataframe`, `dedup`, `summarize_merged_dataframes`, and expects subclasses to implement `cast_types`, `data_columns`, `operations` & `unique_index_columns`, as well as an appropriate `group` method.

---

Actual dataframe library classes are:

* `DataframeDataCollectionStatus`: `'data_collection_status.csv'`
* `DataframeHostMetric`: `'host_metric.csv'`
[...]

PR #285, opened 2025-11-25
cli - use library.dataframes

himdel:
Issue: AAP-54468 / AAP-57129

This is the second part of #237 - use library.dataframes in the CLI.


himdel:
Closing, and linking to #540

PR #288, opened 2025-11-26
Add total count into service collectors for events and jobhostsummary

MilanPospisil:
[AAP-59316]

Adds counting of total events and jobhostsummaries.

Also renaming some count variables to more unified shape.

### DIFF
--- a/metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
+++ b/metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
@@ -88,8 +88,8 @@ class EventModulesAnonymizedRollup(BaseAnonymizedRollup):
         Override merge to handle the new structure with event_total and task_summary.
         Concatenates task_summary dataframes and sums event_total.
         """
-        # Handle initial empty dict case
-        if not data_all:
+        # Handle initial None case (first iteration from load_anonymized_rollup_data)
+        if data_all is None:
             return data_new
 
         # Concatenate task_summary dataframes and sum event_totals
@@ -229,6 +229,13 @@ class EventModulesAnonymizedRollup(BaseAnonymizedRollup):
 
         data is a dict with 'event_total' and 'task_summary' dataframe
         """
+
+        # Handle None input (no data files)
+        if data is None:
+            return {
+                'json': {'event_total': 0},
+                'rollup': {'aggregated': pd.DataFrame(), 'event_total': 0},
+            }
 
         # Extract event_total and task_summary dataframe from the data structure
         event_total = data.get('event_total', 0)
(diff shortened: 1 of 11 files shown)


metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -83,10 +83,32 @@ def __init__(self):
...
+
+        # Handle initial empty dict case
+        if isinstance(data_all, dict) and not data_all:
+            return data_new
+
+        # Concatenate task_summary dataframes and sum event_totals
+        return {
+            'event_total': data_all['event_total'] + data_new['event_total'],
+            'task_summary': pd.concat([data_all['task_summary'], data_new['task_summary']], ignore_index=True),
+        }
```
himdel: Is the 

```
        if isinstance(data_all, pd.DataFrame) and data_all.empty:
            return data_new
```

bit still functional? It seems the inputs are always a dict (a dict with a number, and a dataframe inside)

metrics_utility/anonymized_rollups/jobhostsummary_anonymized_rollup.py
```diff
@@ -10,6 +12,25 @@ def __init__(self):
...
+            return data_new
+
+        # Handle initial empty dict case
+        if isinstance(data_all, dict) and not data_all:
+            return data_new
+
+        # Concatenate aggregated dataframes and sum jobhostsummary_totals
+        return {
+            'jobhostsummary_total': data_all['jobhostsummary_total'] + data_new['jobhostsummary_total'],
+            'aggregated': pd.concat([data_all['aggregated'], data_new['aggregated']], ignore_index=True),
```
himdel, line 27: (same here, can data_all be a dataframe or is that bit dead code?)

himdel (approved):
LGTM, except for the possibility of data_all being a dataframe (rather than a dict consistent with the result of merge)

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
MilanPospisil: Nice catch, it should be handled differently.

metrics_utility/anonymized_rollups/jobhostsummary_anonymized_rollup.py
MilanPospisil: Yes, I believe its dead code.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
MilanPospisil: This is dead code, below is the correct code, this should be deleted.

PR #296, opened 2025-12-04
Milan fill perf testing data

MilanPospisil:
[AAP-54264]

## Description

3 main files:
clean_all_data.py                  Cleans the database
fill_perf_db_data.py              Inserts data into database
run_perf.py                            Runs the final perf test using the generated data

Additionaly, added ability to save packed data in tarballs, or save only csv plain files.

## Testing

cd tools/anonymized_db_perf_data

Your first step, call it for first time and thereafter whenever you want to clean data:
./clean_all_data.py               

Filling the data (adjust params as needed):
./fill_perf_db_data.py --host-count 5000 --job-count 100

Running the perf test using prepared data:
./run_perf.py

You can see inside out folder anonymized.json with computed results. And also rollups with some unanonymized computed data.

The fill perf db data can be called multiple times, each time it adds new data.

You can easily run off the disk. For proper large data testing, we need around 50 GB of data  (mine laptom dont have so much free space anymore...).


ShaiahWren (dismissed):
What do you think about adjusting run() function in tools/anonymized_db_perf_data/helpers.py on line 50 to accept a params parameter for parameterized queries? The SQL injection via f-string interpolation is a security concern.   

See precedent in [metrics_utility/library/collectors/controller/main_jobevent_service.py:27](https://github.com/ansible/metrics-utility/blob/983daa97722a49d34752385e25ed03277541931d/metrics_utility/library/collectors/controller/main_jobevent_service.py#L27):
  cursor.execute(jobs_query, {'since': since, 'until': until}) as well as [metrics_utility/library/lock.py:21](https://github.com/ansible/metrics-utility/blob/983daa97722a49d34752385e25ed03277541931d/metrics_utility/library/lock.py#L21)
  
 These functions that need updating to use the parameterized queries:
 - create_organization() 
  - create_inventory() 
  - create_project() 
  - create_job_templates() 
  - create_hosts() 
  - create_job() 
  - create_job_host_summaries() 
  - create_job_events() 
  - create_jobevent_partitions()

ShaiahWren:
Consider adding transaction management to ensure atomicity for multi-step operations:

  In `fill_perf_db_data.py`:
  - fill_init_data() - Lines 58-70
  - fill_job() - Lines 139-141

  In `helpers.py`:
  - delete_all() - Lines 203-214

  Currently if any operation fails mid-way, previous operations remain
  committed, leaving orphaned data in the database.

ShaiahWren:
Hi again Milan :)
Should the amount of data be larger (more realistic) - (100-200 jobs, ~200K events)?

Also: The OLD collectors filter by main_jobhostsummary.modified (currently Jun 2025+) but jobs have created dates in Jan 2024. This causes OLD collectors to return 0 rows while NEW collectors work fine.

  Can you update fill_perf_db_data.py to set modified dates that align with job creation dates? Need this to fairly compare OLD vs NEW performance.

 Thanks!

ShaiahWren (approved):

MilanPospisil:
> What do you think about adjusting run() function in tools/anonymized_db_perf_data/helpers.py on line 50 to accept a params parameter for parameterized queries? The SQL injection via f-string interpolation is a security concern.
> 
> See precedent in [metrics_utility/library/collectors/controller/main_jobevent_service.py:27](https://github.com/ansible/metrics-utility/blob/983daa97722a49d34752385e25ed03277541931d/metrics_utility/library/collectors/controller/main_jobevent_service.py#L27): cursor.execute(jobs_query, {'since': since, 'until': until}) as well as [metrics_utility/library/lock.py:21](https://github.com/ansible/metrics-utility/blob/983daa97722a49d34752385e25ed03277541931d/metrics_utility/library/lock.py#L21)
> 
> These functions that need updating to use the parameterized queries:
> 
> * create_organization()
> * create_inventory()
> * create_project()
> * create_job_templates()
> * create_hosts()
> * create_job()
> * create_job_host_summaries()
> * create_job_events()
> * create_jobevent_partitions()

This would be indeed security concern in normal app, but since this is script running by developers and the strings inside SQL are generated by the scripts, we are ok :-) Whoever has access to the script has access also to the DB itself, so SQL fixing injection will prevent nothing :-)

MilanPospisil:
> Consider adding transaction management to ensure atomicity for multi-step operations:
> 
> In `fill_perf_db_data.py`:
> 
> * fill_init_data() - Lines 58-70
> * fill_job() - Lines 139-141
> 
> In `helpers.py`:
> 
> * delete_all() - Lines 203-214
> 
> Currently if any operation fails mid-way, previous operations remain committed, leaving orphaned data in the database.

Orphaned records are fine in this case, because this is for performance tests only. We do not care if for example job has no events, it does not matter for our purposes :-) If this would be part of the production, yes, transactions will be very important.

But if you will add transactions into this script, you will:

1) Gain very little software reliability, since we dont care about corrupted data (unless it will in rare cases lead to failures, but we can also clean all data and fill them again)

2) Transactions will add execution time overhead, so script will be slower for filling the data - and its pain even now :-)

MilanPospisil:
> Hi again Milan :) Should the amount of data be larger (more realistic) - (100-200 jobs, ~200K events)?
> 
> Also: The OLD collectors filter by main_jobhostsummary.modified (currently Jun 2025+) but jobs have created dates in Jan 2024. This causes OLD collectors to return 0 rows while NEW collectors work fine.
> 
> Can you update fill_perf_db_data.py to set modified dates that align with job creation dates? Need this to fairly compare OLD vs NEW performance.
> 
> Thanks!

You can set up number of jobs and events in the script arguments and if you will run filling the data multiple times, it will stack, so you can for sure create much more data. The script without parameters is more like for testing purposes that it works, so it ends in reasonable short time.

> Also: The OLD collectors filter by main_jobhostsummary.modified (currently Jun 2025+) but jobs have created dates in Jan 2024. This causes OLD collectors to return 0 rows while NEW collectors work fine.
> 
> Can you update fill_perf_db_data.py to set modified dates that align with job creation dates? Need this to fairly compare OLD vs NEW performance.

I dont understand this one? You can run both old code and new code (once we will have it, we now have only new code) on the same data - and it is also reccommended, if you will have different data sets for new and old data, results will not be representable.

PR #313, opened 2026-01-08
library.dataframes.BaseTraditional - move merge logic to child classes

himdel:
Updates the library traditional dataframes to move the full merge logic into a method in each of the 3 classes.

`merge` was a shared method, calling `summarize_merged_dataframes` & `cast_datafame`, also shared, both called with details from per-class `unique_index_columns`, `data_columns` & `cast_types`

the abstraction makes sense, but we can remove extra complexity with it, and the result is a simple set of pandas instructions. 

(Also drops the logic casting all indexes to objects.)

wip - test with #285


himdel:
Closing, and linking to https://github.com/ansible/metrics-utility/pull/540

PR #318, opened 2026-01-21
pytest workflow: drop per-PR coverage reporting

himdel:
Drops per-PR coverage reporting from per-PR runs, while still doing it on merge.
No need for pull_request_target, changed back to pull_request.

(Followed by #321)


MilanPospisil (approved):
LGTM :+1: 

But why we are dropping the sonar cloud analysis per pr, is it useless in your opinion?

PR #321, opened 2026-02-05
[AAP-63367] Remediate PwnRequest Vulnerability in GitHub Actions Workflows

pmflanagan:


.github/workflows/sonar_checks.yml
```diff
@@ -0,0 +1,89 @@
...
+permissions: read-all
+
+jobs:
+  sonarcloud:
+    name: SonarCloud Static Scans (PR)
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request'
+
+    steps:
+      - name: Checkout Code
```
himdel, line 19: we probably shouldn't be checking out the repo at all in this one?

or, if we do, do we need both the main branch (here), and the PR branch (line 67)?

.github/workflows/pytest.yml
```diff
@@ -9,16 +9,10 @@ jobs:
   pytest:
     runs-on: ubuntu-latest
     steps:
-      - name: "Checkout metrics-utility (${{ github.ref }})"
+      - name: "Checkout metrics-utility"
         uses: actions/checkout@v6
-        if: ${{ ! github.base_ref }}
-
-      - name: "Checkout metrics-utility (${{ github.event.pull_request.head.ref }})"
```
himdel, line 16: this doesn't look like a valid change ... both checkouts are still needed to run the tests, as they were, right?

or do they end up doint the same thing now that pull_request is used again?

.github/workflows/pytest.yml
```diff
@@ -95,6 +89,23 @@ jobs:
         run: |
           uv run pytest -s -v --cov=. --cov-report=xml
 
+      - name: "Upload code coverage report"
+        uses: actions/upload-artifact@v4
```
himdel, line 94: these upload steps should only happen for the PR case, right?

Missing `if: ${{ github.base_ref }}`

.github/workflows/pytest.yml
pmflanagan: The single checkout is required for the security fix. With `pull_request`, that one checkout still runs the tests on the PR (merge/head) as intended

.github/workflows/pytest.yml
pmflanagan: Thanks for pointing that out, I'll update this in another commit.

himdel: ooh, I now see `if: github.event_name == 'pull_request'` with the others .. shoudld we use the same here?

himdel (approved):

.github/workflows/sonar_checks.yml
pmflanagan: Both are needed. 

The first gives the repo and workspace so we can run `actions/download-artifact` and `gh pr checkout`. Otherwise we don't have a repo to work in.

The second put's the code in the working tree so Sonar analyses the correct branch. Coverage artifact stays in `coverage-artifact` so it doesn't get overwritten.

PR #322, opened 2026-02-09
ruff updates

himdel:
looks like new version of ruff is out and the formatting rules changed slightly - updating
bump the version of ruff used by pre-commit to match uv.lock.
and drop python 3.11 from sonar config, leave only 3.12


MilanPospisil (approved):
LGTM 👍

PR #323, opened 2026-02-09
add --date flag to generator

ShaiahWren:
[AAP-64856]

## Description

The generator script fill_perf_db_data.py is hard-coded to generate data across the month of January 2024. For performance testing, we need the option to target one day with this script to ensure data covers most hours in the 24 hour period we are testing. Enter the `--date` param.

Usage example: 
```
.venv/bin/python tools/anonymized_db_perf_data/fill_perf_db_data.py \
      --date=2024-01-25 \
      --job-count=20 \
      --host-count=1000 \
      --task-count=50
```

### Steps to Test

1.
2.
3.

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

## Notes for Reviewers

### DIFF
--- a/metrics_utility/test/conftest.py
+++ b/metrics_utility/test/conftest.py
@@ -38,7 +38,7 @@ def setup_processed_dataframe(fixed_now):
         mock_df_for_batch_processed[col] = pd.to_datetime(mock_df_for_batch_processed[col], utc=True).dt.tz_localize(None)
 
     mock_df_for_batch_processed['last_deleted'] = mock_df_for_batch_processed['last_deleted'].apply(
-        lambda x: (x.isoformat(timespec='seconds') if pd.notna(x) and isinstance(x, dt_actual.datetime) else None)
+        lambda x: x.isoformat(timespec='seconds') if pd.notna(x) and isinstance(x, dt_actual.datetime) else None
     )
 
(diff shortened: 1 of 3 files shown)


himdel (changes requested):
The changes as they are look good 👍 , just, not enough changes :)...

Just try running with `--date=2025-01-01` or such.

The problem is, `JOB_DATE_START` and `JOB_DATE_END` are not the only places with a hardcoded datetime - there's also `create_jobevent_partition`.

`create_jobevent_partition` is called first to create the partition tables needed so we can actually add the data to them - but it just iterates through all hours between 2024-01-01 and 2024-02-01, ignoring the 2 variables .. fix that and I think this should work :)

---

Ideally, also, setting some constants and then changing them from outside is not an ideal pattern.. and we don't really need to :) The only places using them are `get_job_timestamps` (which is called by `create_job`) and `create_jobevent_partitions`. Both `create_job` and `create_jobevent_partitions` are called from `fill_perf_db_data.py` - so main could just be setting the 2 dates there and passing them as params.

himdel (approved):
Works for me 👍 :)

(I just had to rebase to sync with the new schema from #300, but that will take care of itself when merged :).)

PR #332, opened 2026-02-24
Fix venv re-activation in perf tools

himdel:
Prevent re-activating virtualenv when already running inside one - by checking `VIRTUAL_ENV` before activation.

`clean_all_data.py`: accept 'y' as confirmation too


MilanPospisil (approved):
LGTM 👍

PR #340, opened 2026-03-03
coderabbit: move summary from PR description to first comment

himdel:
inherit coderabbit org config,
but change it to NOT edit PR description

(see also ansible/metrics-service#116 - setting a different `high_level_summary` value here, to see what does what)

## Summary by CodeRabbit

* **Chores**
  * Added configuration file to enable high-level summary reviews and inheritance behavior in the code review process.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,4 @@
+inheritance: true
+reviews:
+  high_level_summary: true
+  high_level_summary_in_walkthrough: true


MilanPospisil (approved):
LGTM 👍

MilanPospisil:
How does this settings ensures "change it to NOT edit PR description" ?

himdel:
Well, I hope it does :) https://docs.coderabbit.ai/reference/configuration

It worked in https://github.com/ansible/metrics-service/pull/116, where we're setting false & true, which stops the summary completely, and doesn't put it in the first comment.

Sherin wanted to keep the summary, just as part of the first comment, so .. maybe this combination does that :)
If not, we'll just change both to `high_level_summary: false` only :)

PR #344, opened 2026-03-16
Milan hardening3

MilanPospisil:
[AAP-[66352](https://issues.redhat.com/browse/AAP-66352)]

Adds some extra tests, but mainly it overwrittes django basic JSON parsing for our cursor, now cursor is returning dict of json instead of parsable text. It should override only cursor behavior, not the global behavior.

Anonymized rollup now expects only dict for installed collections in unified jobs.

This is important perf boost, because:

1) There are too many jobs and number of installed collections is large and also grows with how large customer is, this makes our time complexity N*M, which does not scale well (N = jobs, M = collections)

2) Previous optimization used cache for already parsed json  (compares hash of json with cache), but it seems that jsonb does not guarantee the same json for the same jsonb (keys order can vary and such...)

### DIFF
--- a/metrics_utility/anonymized_rollups/jobs_anonymized_rollup.py
+++ b/metrics_utility/anonymized_rollups/jobs_anonymized_rollup.py
@@ -1,6 +1,3 @@
-import hashlib
-import json
-
 import pandas as pd
 
 from metrics_utility.anonymized_rollups.base_anonymized_rollup import BaseAnonymizedRollup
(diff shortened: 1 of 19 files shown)


MilanPospisil:
@coderabbitai review

MilanPospisil:
Do not merge until tech preview release.

himdel (changes requested):
The removals of the extra hashing logic look good,
the jsonb change - not so much.

This is not a performance optimization, not unless you have real numbers, most likely the json.load implementation is the same one in both cases.

And it means we now have to deal with every field possibly being structured data. Is it worth it? Does it give us anything?

I'd much rather see explicit typing for each field we get from the db, rather than implicit transforms that can affect anything.

MilanPospisil:
> The removals of the extra hashing logic look good, the jsonb change - not so much.
> 
> This is not a performance optimization, not unless you have real numbers, most likely the json.load implementation is the same one in both cases.
> 
> And it means we now have to deal with every field possibly being structured data. Is it worth it? Does it give us anything?
> 
> I'd much rather see explicit typing for each field we get from the db, rather than implicit transforms that can affect anything.

Hmmm... Im reading jsonb is transformed into plain string - thats true, I was thinking more like some binary that can transform into python struct more faster, ... so Im closing this PR.

PR #350, opened 2026-03-17
Hardening of anonymized data part 3

MilanPospisil:
[AAP-[66352](https://issues.redhat.com/browse/AAP-66352)]

1) Controller version was missing filter - it was returning also execution node versions, which contained ansible version, not AWX version.

2) The caching of json based on string hash is not super reliable - it relyies on implementation detail of postgres jsonb serialization, which sorts keys, which produces the same text for the same data (with different order of keys), but this is not API contract and may change. So we now build more reliable caching based on execution environment id - we are only caching installed collections based on this id, because the same EE id, the same installed collections. Of course, this may little change, when between two job runs with the same EE will be some collection installation, but those situations are neghligble in big data. Json string cache is used as fallback when EE id is missing.

3) Finished jobs are now filtered correctly - previous version worked magicaly based on pandas version - it was either copy, or view, view can then silently fails on adding new columns.


MilanPospisil:
@coderabbitai review

cshiels-ie (approved):

metrics_utility/anonymized_rollups/jobs_anonymized_rollup.py
```diff
@@ -650,15 +653,20 @@ def _process_collections_dict(self, collections_data, collections_stats, failed,
...
+        Return a hashable cache key for the installed_collections of a row.
 
-        Used as a cache key to avoid re-parsing identical JSON payloads
-        (e.g. all jobs that share the same execution environment).
-        SHA-256 is used instead of the raw string to keep the cache memory-efficient
-        when the JSON payload is large.
+        Prefers execution_environment_id as the key because it is a stable
+        integer that uniquely identifies a fixed collection set.  Prefixes the
+        tuple with 'ee' to avoid any collision with the fallback hash keys.
+        Falls back to hashing the raw string for rows that have no EE id.
```
himdel, line 663: The only rows without EE id would be rows from previous versions, right?
Might be a good idea to drop the workaround soon, 
but probably makes sense to keep for now :+1:

himdel (approved):

metrics_utility/anonymized_rollups/jobs_anonymized_rollup.py
MilanPospisil: AWX should enforce EEs for newever versions, but Im not sure if some customers have some workaround or even older versions... maybe we can remove the fallback and nothing will happen :-)

Older versions have EE only as obligatory...

And for very old AWX versions, it will fail anyway :-( Because ee_id field will not be in db! But we are not supporting those very old installations.

But is it something that we should be solving? What do you think? Service is anyway installed only in recent versions?

Lets merge it and we will see later :-)

PR #363, opened 2026-04-01
AAP-68113 [PENDING] Candle pin v2

cshiels-ie:
# Candlepin Integration — AAP-73765

## Overview

Adds Candlepin consumer identity certificate support to metrics-utility, enabling AAP Controller instances to authenticate analytics uploads to \`cert.console.redhat.com\` using mTLS instead of a service account.

The Controller (AWX PR #16388) owns the Candlepin lifecycle — registration, check-in, and renewal. On each gather run, metrics-utility reads the cert directly from the AWX \`conf_setting\` database table and uses it without running a parallel lifecycle. If the Controller DB has no cert (e.g. standalone deployments without AWX PR #16388), metrics-utility falls back to its own local store and lifecycle.

When a valid Candlepin certificate is present the upload path switches to mTLS automatically. If the certificate is absent, expired, or the TLS handshake fails, the system falls back to service-account authentication so existing deployments are never broken.

> **Depends on:** #431 (NaN/None handling fixes in dedup/dataframe layers — must merge first)

---

### \`metrics_utility/library/candlepin/client.py\` — \`CandlepinClient\`

Thin REST client for the Candlepin API. All API calls that require authentication after initial registration use the consumer identity certificate (mTLS), matching the pattern used by \`subscription-manager\`.

| Method | Endpoint | Auth | Purpose |
|---|---|---|---|
| \`discover_org\` | \`GET /users/{username}/owners\` | Basic | Discover the Candlepin org key |
| \`register_consumer\` | \`POST /consumers?owner={org}\` | Basic | Register this AAP instance and obtain an identity certificate |
| \`checkin\` | \`PUT /consumers/{uuid}\` | mTLS | Reset the inactivity timer; best-effort, never raises |
| \`regenerate_cert\` | \`POST /consumers/{uuid}\` | mTLS | Force certificate renewal; raises \`RuntimeError\` on failure |

TLS verification defaults to \`True\` (system CA store). Pass \`candlepin_ca\` to verify against a specific CA bundle. Temporary PEM files written to disk for mTLS calls are created with \`0o600\` permissions and unconditionally deleted on context-manager exit.

### \`metrics_utility/library/candlepin/lifecycle.py\`

Certificate inspection and lifecycle orchestration helpers.

| Function | Purpose |
|---|---|
| \`parse_cert(pem_text)\` | Parse a PEM cert; return serial, CN, expiry, days remaining, etc. |
| \`is_cert_valid(cert_pem)\` | Guard check: parseable, not-yet-valid, not-expired. |
| \`needs_renewal(pem_text, days_before_expiry)\` | \`True\` if the cert expires within the threshold. |
| \`get_candlepin_ca()\` | Returns CA bundle path: env var → auto-detected RHSM CA → None. |
| \`run_candlepin_lifecycle(...)\` | Orchestrate one gather-run lifecycle: inspect cert → check-in → proactive renewal if needed. |

### \`metrics_utility/library/candlepin/store.py\`

Storage abstraction with two backends selected via \`METRICS_UTILITY_CANDLEPIN_STORAGE\`:

- \`LocalCandlepinStore\` (default) — reads/writes \`cert.pem\`, \`key.pem\`, \`uuid.txt\` on the local filesystem with atomic writes and \`0o600\` permissions.
- \`DBCandlepinStore\` — reads cert/key/UUID from the AWX \`conf_setting\` table using the AWX ORM and \`decrypt_field()\` to handle AWX-encrypted values. Key names match what AWX PR #16388 writes: \`CANDLEPIN_CERT_PEM\`, \`CANDLEPIN_KEY_PEM\`, \`CANDLEPIN_CONSUMER_UUID\`.

> **Note on \`DBCandlepinStore\` and encryption:** \`CANDLEPIN_CERT_PEM\` and \`CANDLEPIN_KEY_PEM\` are stored encrypted in \`conf_setting\` (AWX \`encrypted=True\`). Decryption requires AWX's \`SECRET_KEY\` via \`decrypt_field(setting, 'value')\`. This means \`DBCandlepinStore\` only works when running inside the AWX Django context (i.e. the \`automation-controller-task\` container). A future standalone metrics-service would need a Controller API endpoint rather than direct DB access.

---

### \`metrics_utility/management/validation.py\`

**\`handle_crc_ship_target\` execution order**

1. **Check controller DB first** — call \`_load_cert_from_controller_db()\`, which reads and decrypts from AWX \`conf_setting\` via \`DBCandlepinStore\` and validates with \`is_cert_valid()\`.
2. If a valid cert is found in the controller DB → use it directly, skip the metrics-utility lifecycle (the Controller owns check-in and renewal). Return immediately.
3. If no cert in controller DB (DB unavailable, empty, or cert expired) → fall through to the local store path:
   - Load from the configured local store.
   - If local store is empty, seed it from the controller DB (one-time seed).
   - If still no cert and registration is enabled (default: on) → attempt standalone registration.
   - If cert available and lifecycle is enabled (default: on) → check-in + proactive renewal.
4. Inject \`candlepin_cert_pem\` / \`candlepin_key_pem\` into \`billing_provider_params\` for the upload step.

All steps are best-effort: failures log a warning or error and fall through to service-account auth.

### \`metrics_utility/automation_controller_billing/package/package_crc.py\`

- \`shipping_auth_mode()\` returns \`SHIPPING_AUTH_CERTIFICATES\` when a valid Candlepin cert is present, \`SHIPPING_AUTH_SERVICE_ACCOUNT\` otherwise.
- \`ship()\` manages the temp-file lifecycle for mTLS uploads. On \`SSLError\`, falls back to service-account auth if credentials are configured; raises \`FailedToUploadPayload\` if not.
- \`_send_data()\` gains a \`SHIPPING_AUTH_CERTIFICATES\` branch that posts to \`cert.{hostname}\` with \`session.cert\` and server-cert verification.
- \`_gather_config()\` (via \`collector.py\`) strips \`candlepin_cert_pem\` and \`candlepin_key_pem\` from the payload archive — key material is never uploaded.

---

### \`uv run ./manage.py candlepin_manage\`

```
subcommands:
  register   Register this instance as a Candlepin consumer
  renew      Check in and renew the Candlepin identity certificate
```

Global options (e.g. \`-v\`/\`--verbosity\`) must be placed **before** the subcommand:
- \`manage.py candlepin_manage -v 2 renew\` ✓
- \`manage.py candlepin_manage renew -v 2\` ✗

---

## New environment variables

| Variable | Default | Purpose |
|---|---|---|
| \`METRICS_UTILITY_CANDLEPIN_URL\` | \`https://subscription.rhsm.redhat.com/subscription\` | Candlepin base URL |
| \`METRICS_UTILITY_CANDLEPIN_CA\` | _(auto-detected from \`/etc/rhsm/ca/\` if present, else system CA store)_ | Path to Candlepin CA cert for TLS verification |
| \`METRICS_UTILITY_CANDLEPIN_RENEWAL_DAYS\` | \`30\` | Days before expiry at which proactive renewal is triggered |
| \`METRICS_UTILITY_CANDLEPIN_STORAGE\` | \`local\` | Store backend: \`local\` (filesystem) or \`db\` (AWX conf_setting) |
| \`METRICS_UTILITY_CANDLEPIN_CERT_DIR\` | \`/var/lib/awx/candlepin-certs\` | Directory for local cert storage |
| \`METRICS_UTILITY_CANDLEPIN_REGISTRATION_ENABLED\` | \`true\` | Disable standalone consumer registration on gather runs (\`false\` to opt out) |
| \`METRICS_UTILITY_CANDLEPIN_LIFECYCLE_ENABLED\` | \`true\` | Disable check-in and proactive renewal on gather runs (\`false\` to opt out) |

---

## Fallback chain

```
gather run (SHIP_TARGET=crc)
  │
  ├─ 1. Controller DB has valid cert?
  │      YES → mTLS upload (no metrics-utility lifecycle)
  │      NO  ↓
  │
  ├─ 2. Local store has cert?
  │      YES → metrics-utility lifecycle (check-in + renewal) → mTLS upload
  │      NO  ↓
  │
  ├─ 3. Seed from controller DB possible?
  │      YES → save to local store → metrics-utility lifecycle → mTLS upload
  │      NO  ↓
  │
  ├─ 4. Registration enabled?
  │      YES → register → save to local store → metrics-utility lifecycle → mTLS upload
  │      NO  ↓
  │
  └─ 5. No cert → service-account auth (OIDC / basic)
         On SSLError during mTLS → also falls back here
```

---

## Dependencies

\`cryptography>=44.0.1\` added to \`pyproject.toml\` for X.509 certificate parsing. Lower bound 44.0.1 excludes CVE-2024-12797 and the 2025 bundled-OpenSSL vulnerability.

---

## AI Assistance
This change was developed with assistance from Claude Code (Claude Sonnet 4.6).


D4D3VD4V3 (commented):
Reviewing

metrics_utility/management/validation.py
```diff
@@ -164,12 +189,164 @@ def handle_not_s3():
...
+        'INSTALL_UUID',
+    ]
+    try:
+        from django.db import connection
+
+        placeholders = ', '.join(['%s'] * len(keys))
+        query = 'SELECT key, value FROM conf_setting WHERE key IN (' + placeholders + ')'
+        with connection.cursor() as cursor:
+            cursor.execute(query, keys)
+            rows = {key: json.loads(value) for key, value in cursor.fetchall() if value}
```
D4D3VD4V3: This block reads `REDHAT_PASSWORD` / `SUBSCRIPTIONS_PASSWORD` via raw SQL + `json.loads()`. AWX stores sensitive settings encrypted (`$encrypted$UTF8$AESCBC$...`), so `json.loads()` returns the ciphertext string rather than the plaintext password.

`DBCandlepinStore.load()` was explicitly fixed in `0aad029` to use `Setting.objects.filter()` + `decrypt_field()` for exactly this reason, but this credential fetcher was never updated to match.

If these password fields are encrypted (likely in production AWX), Candlepin registration silently fails with an opaque auth error. Consider using the same `Setting.objects.filter()` + `decrypt_field()` pattern here.

D4D3VD4V3: Only thing I could find with Claude. Went through the rest of the code and tested it E2E and in isolation, and can see that it picks up controller candlepin cert if it exists or uses its own. LGTM!

D4D3VD4V3 (approved):
Only small feedback with decrypting AWX field in validation.py. Otherwise looks great. Thanks Ciaran!

cshiels-ie:
Fixed in cde0e44 — `_fetch_registration_credentials_from_db()` now uses `Setting.objects.filter()` + `decrypt_field()` / `is_encrypted()` identically to how `DBCandlepinStore.load()` handles the cert fields. Removed the now-unused `import json` from `validation.py` as well.

Note: `DBCandlepinStore.load()` had the same raw-SQL pattern for the cert/key fields, which the fork had already fixed in 0aad029. This commit only adds the equivalent fix for the credential path.

PR #371, opened 2026-04-22
Milan change installed collections name to collection

MilanPospisil:
[AAP-72552]

## Description

Rename installed collections "name" to "collection", so segment will not anonymize it.

### DIFF
--- a/metrics_utility/anonymized_rollups/anonymized_rollups.py
+++ b/metrics_utility/anonymized_rollups/anonymized_rollups.py
@@ -24,6 +24,29 @@ def hash(value, salt):
     return hashed
 
 
+def _installed_collection_name_is_unknown(collection_name: Any, known: Dict[str, Any]) -> bool:
+    """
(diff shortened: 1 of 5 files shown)


himdel (changes requested):
This seems to be changing the format of the data we're sending to segment after the release.

Furthermore, are we just assuming that `name` is magic in segment?
I think it's more likely that someone explicitly set a rule to transform that field in segment - we should not be writing code to work around that.

MilanPospisil:
> This seems to be changing the format of the data we're sending to segment after the release.
> 
> Furthermore, are we just assuming that `name` is magic in segment? I think it's more likely that someone explicitly set a rule to transform that field in segment - we should not be writing code to work around that.

https://www.twilio.com/docs/segment/privacy/portal

Here, it seems name is by default anonymized and can not be turned off.

himdel (approved):

PR #372, opened 2026-04-23
Remove use bulk from split into segment

MilanPospisil:
# [AAP-72806] 

## Description

What it fixes - our Segment does not support use bulk mode, I removed it from the service and utility.

### DIFF
--- a/metrics_utility/library/storage/segment.py
+++ b/metrics_utility/library/storage/segment.py
@@ -17,17 +17,14 @@ except ImportError:
 
 
 class StorageSegment:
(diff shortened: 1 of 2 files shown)


himdel (commented):
LGTM,
but needs a corresponding metrics-service PR merged first, otherwise it breaks there

MilanPospisil:
> LGTM, but needs a corresponding metrics-service PR merged first, otherwise it breaks there

Im aware, thanks for putting there note :-)

cshiels-ie (approved):

PR #395, opened 2026-05-04
Remove unused django-ansible-base dependency

himdel:
Closes #364 

The `django-ansible-base` dependency is not used in metrics-utility (unlike metrics-service) since #259 replaced it with `library.lock`.


MilanPospisil (approved):
LGTM 👍

PR #397, opened 2026-05-04
_calculate_collection_interval: reuse max gather period days, and timedelta

himdel:
Cosmetic change,
compute the 2 values once, rather than for every logging,
and update logging to always report a change, but reduce the value to info/debug when logging something that's not a user error.

Thus...

|what|old|new|
|-|-|-|
| Original since-until | warning | debug |
| End in future | warning | warning |
| Since in future | warning | warning |
| Too long interval | warning | warning |
| Since set, until unset | - | info |
| Since unset, until unset | - | info |
| Since after until | warning | warning |
| Too long interval | warning | warning |
| Ignoring last run | warning | warning |
| Final since-until | warning | info |

### DIFF
--- a/metrics_utility/base/collector.py
+++ b/metrics_utility/base/collector.py
@@ -136,9 +136,12 @@ class Collector:
     #
     def _calculate_collection_interval(self, since, until):
         _now = now()
+        _max = get_max_gather_period_days()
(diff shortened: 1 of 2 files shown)


himdel:
Changes applied in the gather-only branch

PR #400, opened 2026-05-13
exclude non-source dirs from sonar scanning

himdel:
adds a `sonar.exclusions`, explicitly excluding:

* `tools/` - dev env tooling, perf test tooling, testathon data generators
* `workers` - old library examples
* `.github/` - workflow tooling
* `mock_awx` - tests mock
* `docs` - docs
* `sql` - a sql dump
* `run-*` - manual testing scripts


himdel:
Already that way because `sonar.sources`

PR #419, opened 2026-05-28
tests run_*_ext: don't create __pycache__, remove unused var, add lang & tz default

himdel:
During tests, `run_build_ext` and `run_gather_ext` run the full command in a new subprocess, with clean environment (except adding `AWX_LOGGING_MODE=stdout`). The same happens in snapshot tests.

Which means running tests leaves `__pycache__` dirs all over the tree, ignoring user settings like `PYTHONPYCACHEPREFIX` or `PYTHONDONTWRITEBYTECODE`.

Additionally, `AWX_LOGGING_MODE=stdout` is only relevant when running in the actual controller environment, which is not something happening during the tests - removing.

This changes the clean environment to:

```
LANG=en_US.UTF-8
PYTHONDONTWRITEBYTECODE=1
TZ=UTC
```

### DIFF
--- a/metrics_utility/test/snapshot_tests/snapshot_utils.py
+++ b/metrics_utility/test/snapshot_tests/snapshot_utils.py
@@ -1,4 +1,3 @@
-import copy
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


MilanPospisil (approved):
LGTM 👍

PR #428, opened 2026-06-08
test: add JSON schema + tests to validate with billing

hunterkepley:
- Add JSON schemas for config (v1.0 and v2.0), data_collection_status, job_host_summary, and total_workers_vcpu
- Expand test_json_schema.py with validation tests for each schema against generated or pre-existing tarball data
- Refactor test helpers to reduce duplication (_load_schema, _generate_and_extract, cleanup_glob fixture)

Issue: AAP-76005

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

https://redhat.atlassian.net/browse/AAP-76005

> [!WARNING]
> **Failure to notify downstream stakeholders (such as the Analytics team) of schema or version changes can result in outages and loss of revenue.**
>
> Please notify stakeholders to disseminate the change correctly, most notably the **Analytics HCC service**.
> A crtical ticket to track the change prior to promotion.
> The change should not be merged until depedancies have been updated.

## Description

- What is being changed?

Adds `/schemas/` (json schemas for files we send to the billing analytics team)
Adds tests to test the schemas against generated files

- Why is this change needed?

To prevent breaking billing when something changes in the files we send them. To catch issues before they affect anything. To provide a way for billing to confirm changes will be fine with their billing analytics processor in HCC

Adds tests+schemas+required reviewers for schema changes, forcing us to adhere to the current schema, or, change the schema and get approval from someone on the Billing Analytics team

### Steps to Test

1. Run the unit test file I added

### Expected Results

The tests will pass when the schema matches our generated files. The tests will fail when the schema is broken

### Code Quality

- [X] Code is properly linted and formatted.
- [X] All tests (existing and new) pass successfully.
- [X] Tests are added or updated as needed.
- [X] Code includes relevant comments for complex sections.
- [X] Documentation is updated where applicable.  

^^^ **I WILL UPDATE THE HANDBOOK LATER AS A SEPARATE PIECE TO THIS TASK/EPIC**


hunterkepley:
test pipeline is failing due to `CODEOWNERS` having 2 members not being members in this repo

CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Schema ownership, should be at least 1 person from Billing Analytics, 
+#   and 1 person from ES approving changes in this directory
+/schemas/ @hunterkepley @himdel @SherinV @cshiels-ie @MilanPospisil @shanemcd @zsadeh @dmzoneill
```
himdel: `Unknown owner on line 3: make sure @zsadeh exists and has write access to the repository`
`Unknown owner on line 3: make sure @dmzoneill exists and has write access to the repository`

sounds like we either need to add them to the write access team (or switch to an action that comments with a ping or someting)

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,199 @@
...
+            try:
+                member = tar.extractfile('./job_host_summary.csv')
+            except KeyError:
+                continue
+            assert member is not None, 'job_host_summary.csv not found in tarball'
+            reader = csv.DictReader(io.TextIOWrapper(member))
+            data = list(reader)
+            if len(data) > 0:
+                schema = _load_schema(request, 'job_host_summary.jsonschema')
+                yield data, schema
```
MilanPospisil: Is the yield necessary there? Would return data and schema will be enough? And its also more readable. If the yield must be there, please at least add comment why.

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,199 @@
...
+
+# This test is different than the others. Using the mock DB, we cannot generate this file due
+# to missing data. To save LoC and make this work with what we already have in place,
+# it uses a pre-existing tarball instead of generating a new one
+existing_data_glob = './metrics_utility/test/test_data/data/*/*/*/*.tar.gz'
+
+
+@pytest.fixture
+def generated_job_host_summary(request):
+    tarballs = glob.glob(existing_data_glob)
```
MilanPospisil: This will return tarballs in different order based on operating system and you are validating only first, thus the test is not stable.

Right now it passes on my local machine, but fails on CI.

The problem in failing CI is that some older tarballs have incorrect schema. Im not sure if its even necessary to validate repo stored tarballs, some of them were even created artificialy, not using collector at all.

Maybe its better to keep only test that calls collector and validate schema against stored tarball.

@himdel Was the existing tarball validation something you explicitly requested?

metrics_utility/test/validation/test_json_schema.py
hunterkepley: I'm happy to have a `return` instead, even though i normally do not, i asked claude to help out with the `generated_<FILE_NAME>` functions. Even though it has kind of the right idea using yield for lazy data generators, in this function specifically it was badly done, almost in a useless way (return right after the yield, resulting in pretty much just a return)

Honestly this is another reason i dont like using AI models on my code 😅 when I miss something dumb like this from it's output, it makes me look bad 😆 

Anyway, tested with just a return instead of the yield, and no issues, so will be adding that change, thanks for the comment

metrics_utility/test/validation/test_json_schema.py
himdel: Sounds like the glob should only look for tarballs created by this test, maybe different dir?

CODEOWNERS
hunterkepley: Removed from this MR

Going with an action, as suggested, in a separate MR to support this rather than a CODEOWNERS file

metrics_utility/test/validation/test_json_schema.py
hunterkepley: there's a comment above this function explaining a bit of this, but ill go into more detail here;

originally, I tried grabbing the new tarballs (with the `uuid` var at the top with a bunch of `0`'s). I dont really know why, but they do not contain a `job_host_summary.csv` with data. This is my first time messing around with metrics utility so im probably just missing something dumb, but:

`existing_data_glob = f'./metrics_utility/test/test_data/data/*/*/*/{uuid}-*.tar.gz'` in place of the current `existing_data_glob` string will result in the following error when this test is ran:

```Failed: job_host_summary.csv with data not found in any tarball```

The directory is correct, its just that the new job host summary files are empty, but the old ones have data

```bash
❯ tar -xzvf 00000000-0000-0000-0000-000000000000-2026-06-01-000000+0000-2026-06-02-000000+0000-6-job_host_summary.tar.gz
x ./job_host_summary.csv
x ./config.json
x ./data_collection_status.csv
x ./manifest.json
❯ ls
00000000-0000-0000-0000-000000000000-2026-06-01-000000+0000-2026-06-02-000000+0000-6-job_host_summary.tar.gz
00000000-0000-0000-0000-000000000000-2026-06-01-000000+0000-2026-06-02-000000+0000-7-main_jobevent.tar.gz
config.json
data_collection_status.csv
job_host_summary.csv
manifest.json
❯ cat job_host_summary.csv
id,created,modified,host_name,host_remote_id,ansible_host_variable,ansible_connection_variable,changed,dark,failures,ok,processed,skipped,failed,ignored,rescued,job_created,job_remote_id,job_template_remote_id,job_template_name,inventory_remote_id,inventory_name,organization_remote_id,organization_name,project_remote_id,project_name
```

hunterkepley: I have ideas for how to push forward, but not sure what's best. Should i be creating data to fill out this file? Should I pick a specific tarball out of the old ones we have? Or is there something a bit 'simpler' or more proper?

hunterkepley: Talked with milan, issue seems to be that the date range it is gathering is incorrect. Updating to be a larger range which captures more mock DB data

MilanPospisil: @hunterkepley Based on our discussion yesterday, the problem was the bad range for gather, so it gathered no data. When you repair it, it should produce several tarballs to validate and we will validate only those, is that true?

schemas/data_collection_status.jsonschema
```diff
@@ -0,0 +1,42 @@
...
+      "since",
+      "until",
+      "file_name",
+      "status",
+      "elapsed"
+    ],
+    "additionalProperties": false,
+    "properties": {
+      "collection_start_timestamp": {
+        "type": "string"
```
himdel: can we add the `format: 'data-time'` to these? (timestamp, since, until)

schemas/data_collection_status.jsonschema
```diff
@@ -0,0 +1,42 @@
...
+      "file_name": {
+        "type": "string",
+        "pattern": "^[a-z][a-z0-9_]*\\.(json|csv)$"
+      },
+      "status": {
+        "type": "string",
+        "enum": ["ok", "failed"]
+      },
+      "elapsed": {
+        "type": "string"
```
himdel: I know in a csv everything is technically a string, but it comes in as an int,
so it might be worth preserving .. is there a sane (and safe) way to make type:integer work?
or just a \d+ pattern?

schemas/job_host_summary.jsonschema
```diff
@@ -0,0 +1,119 @@
...
+      "project_remote_id",
+      "project_name"
+    ],
+    "additionalProperties": false,
+    "properties": {
+      "id": {
+        "type": "string"
+      },
+      "created": {
+        "type": "string"
```
himdel: (same as d_c_s here, can we `format:` the datetimes and `type/pattern` the integers?)

schemas/config-1.0.jsonschema
```diff
@@ -0,0 +1,176 @@
...
+    },
+    "external_logger_type": {
+      "type": ["string", "null"]
+    },
+    "install_uuid": {
+      "type": "string"
+    },
+    "instance_uuid": {
+      "type": "string"
+    },
```
himdel, line 67:
```suggestion
    "install_uuid": {
      "type": "string",
      "format": "uuid"
    },
    "instance_uuid": {
      "type": "string",
      "format": "uuid"
    },
```

schemas/config-1.0.jsonschema
```diff
@@ -0,0 +1,176 @@
...
+    "logging_aggregators": {
+      "type": "array",
+      "items": { "type": "string" }
+    },
+    "pendo_tracking": {
+      "type": "string"
```
himdel: I'm not 100% sure here, but isn't this an enum of "on" and "off"?

schemas/config-1.0.jsonschema
```diff
@@ -0,0 +1,176 @@
...
+      "type": ["integer", "null"]
+    },
+    "license_date": {
+      "type": ["string", "null"]
+    },
+    "license_expiry": {
+      "type": "integer"
+    },
+    "license_type": {
+      "type": "string"
```
himdel: this is en enum too..

`none`, `open`, `UNLICENSED`, `trial`, `enterprise`, `developer` (`awx/awx/main/utils/licensing.py`)

MilanPospisil:
@himdel @hunterkepley 

If we are going to add some additional format restrictions (since everything is string :) ), thats what AI generated (generated by cursor and validated by claude agent), we can pick to implement any or all of them, but it needs another validation field by field to be sure:

[schema_additions.md](https://github.com/user-attachments/files/28838678/schema_additions.md)

schemas/data_collection_status.jsonschema
hunterkepley: Had to define my own validator and extend the default jsonschema one to support this, but it will be in the next commit

schemas/data_collection_status.jsonschema
hunterkepley: using `"pattern"` (regex) for this. There isnt something simple and it would require a second custom validator otherwise

schemas/config-1.0.jsonschema
hunterkepley: looks like it is 🙂 good catch

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,210 @@
...
+        member = tar.extractfile('./data_collection_status.csv')
+        assert member is not None, 'data_collection_status.csv not found in tarball'
+        reader = csv.DictReader(io.TextIOWrapper(member))
+        data = list(reader)
+
+    schema = _load_schema(request, 'data_collection_status.jsonschema')
+    return data, schema
+
+
+def validate_with_datetime(schema, instance):
```
D4D3VD4V3, line 162: There might be an easier way to do this without creating your own validator. 
REquires `pip install jsonschema[format]` pkg
Source: https://python-jsonschema.readthedocs.io/en/v4.20.0/validate/ and https://pypi.org/project/rfc3339-validator/

I tested this Claude-generated script locally and it works as expected
```
from jsonschema import validate, FormatChecker
from jsonschema.exceptions import ValidationError

schema = {
    "type": "string",
    "format": "date-time"
}

print("Test 1: Valid datetime")
try:
    validate("2025-06-12T14:30:00+00:00", schema, format_checker=FormatChecker())
    print("Validation succeeded!")
except ValidationError as e:
    print(f"Validation failed: {e.message}")

print()
print("Test 2: Invalid datetime")
try:
    validate("not-a-valid-datetime", schema, format_checker=FormatChecker())
    print("Validation succeeded!")
except ValidationError as e:
    print(f"Validation failed: {e.message}")

```

D4D3VD4V3 (approved):
LGTM. The only thing I was concerned about was the custom validator for date-time for jsonschema. It seems there are other ways to do it. Left a comment for that in case you're interested in an easier way.

metrics_utility/test/validation/test_json_schema.py
hunterkepley: `pip install jsonschema[format]` leads to no matches unfortunately (but, seems the actual dependency is https://pypi.org/project/rfc3339-validator/)

(**I already have the format pkg within jsonschema so i dont need to pip install that anyway**)

I am concerned about that package though, it has not been updated since 2021, so not sure if using that or keeping our own validator is better 🤔

hunterkepley: though it definitely makes it easier, and good to know that's why `date-time` didnt work for me originally

D4D3VD4V3: Not being updated since 2021 doesn't seem bad to me. It's a fixed validation. Don't think the rules have changed.
And it's installed as `pip install "jsonschema[format]"` with quotes, just to clarify :)

hunterkepley: welllll i meant moreso that there's an automation which checks for the packages being no-longer-maintained and we will run into that (and since metrics service consumes metrics utility (other services may too)) 

each service which imports metrics utility would get that since every pkg in metrics utility will be imported

hunterkepley: for example, Billing (HCC) uses metrics util, i believe controller does too; so 3 maybe 4 that i know of will run into that (might be wrong about controller+HCC, they may just call it via a bash command)

but at the very least this + mservice would

i would prefer a custom validator here, due to this, but im open to other opinions on it

pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "botocore==1.35.96",
     "distro==1.9.0",
     "django==5.2.7",
+    "jsonschema>=4.26.0",
```
himdel: wrong place?

Since jsonschema is only used in tests, it should be in dev dependencies, line 53 and on

(Which might also reduce the cost of using `jsonschema[format]`?)

hunterkepley: thanks, moved to dev deps
that definitely would reduce the number of complaints we get for an out of date pkg, but i see no harm in keeping the simple custom validator, and we have the freedom to expand on it (such as if we had a limit on how far apart dates could be, etc)

schemas/config-1.0.jsonschema
```diff
@@ -0,0 +1,180 @@
...
+      "type": "string",
+      "format": "uuid"
+    },
+    "logging_aggregators": {
+      "type": "array",
+      "items": { "type": "string" }
+    },
+    "pendo_tracking": {
+      "type": "string",
+      "enum": ["on", "off"]
```
himdel: `"detailed"` is apparently allowed too

(both 1.0 and 2.0)

schemas/job_host_summary-1.2.jsonschema
```diff
@@ -0,0 +1,135 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "job_host_summary.jsonschema",
```
himdel, line 3: this one will need a version too, there were previous versions  (=file rename)

I think it's ok if we just have `job_host_summary-1.2.jsonschema` and not support older formats

(or if we do, 1.1 is missing `job_created`, and 1.0 `ansible_connection_variable`, `ansible_host_variable`, `host_remote_id`, `job_created`)

schemas/total_workers_vcpu-1.0.jsonschema
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "total_workers_vcpu.jsonschema",
```
himdel, line 3: lets add the version here too for consistency?  1.0
(leaving just `manifest.json` and `data_collection_status.csv` unversioned?)

schemas/config-2.0.jsonschema
```diff
@@ -0,0 +1,180 @@
...
+      "type": ["string", "null"],
+      "format": "uuid"
+    },
+    "logging_aggregators": {
+      "type": ["array", "null"],
+      "items": { "type": "string" }
+    },
+    "pendo_tracking": {
+      "type": ["string", "null"],
+      "enum": ["on", "off"]
```
himdel:
```suggestion
      "enum": ["on", "off", "detailed", null]
```

schemas/config-2.0.jsonschema
```diff
@@ -0,0 +1,180 @@
...
+    },
+    "license_date": {
+      "type": ["string", "null"]
+    },
+    "license_expiry": {
+      "type": ["integer", "null"]
+    },
+    "license_type": {
+      "type": ["string", "null"],
+      "enum": ["none", "open", "UNLICENSED", "trial", "enterprise", "developer"]
```
himdel:
```suggestion
      "enum": ["none", "open", "UNLICENSED", "trial", "enterprise", "developer", null]
```

(enum validations seem to get tested for all values, including null)

himdel (commented):
Looking good,
couple of small issues running on older tarballs,
otherwise LGTM :)

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,210 @@
...
+
+    for tarball_path in tarballs:
+        with tarfile.open(tarball_path, 'r:gz') as tar:
+            try:
+                vcpu_member = tar.extractfile('./total_workers_vcpu.json')
+            except KeyError:
+                continue
+            assert vcpu_member is not None, 'total_workers_vcpu.json not found in tarball'
+            vcpu = json.load(vcpu_member)
+            schema = _load_schema(request, 'total_workers_vcpu.jsonschema')
```
himdel: (also the version)

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,210 @@
...
+        with tarfile.open(tarball_path, 'r:gz') as tar:
+            try:
+                member = tar.extractfile('./job_host_summary.csv')
+            except KeyError:
+                continue
+            assert member is not None, 'job_host_summary.csv not found in tarball'
+            reader = csv.DictReader(io.TextIOWrapper(member))
+            data = list(reader)
+            if len(data) > 0:
+                schema = _load_schema(request, 'job_host_summary.jsonschema')
```
himdel: (also the version)

schemas/job_host_summary-1.2.jsonschema
hunterkepley: lets go with 1.2 for now and get 1.1 / 1.0 in if needed in the future, if that sounds good to you (to keep this MR scoped to what it is and not require more review on new files 😛 )

metrics_utility/test/validation/test_json_schema.py
```diff
@@ -0,0 +1,214 @@
...
+            except KeyError:
+                continue
+            assert member is not None, 'job_host_summary.csv not found in tarball'
+            reader = csv.DictReader(io.TextIOWrapper(member))
+            data = list(reader)
+            if len(data) > 0:
+                schema = _load_schema(request, f'job_host_summary-{job_host_summary_version}.jsonschema')
+                return data, schema
+
+    pytest.fail(f'job_host_summary-{job_host_summary_version}.csv with data not found in any tarball')
```
himdel: this refers to the csv, so the version is misleading in the name? both

hunterkepley: removed

hunterkepley: also removed from the `total_workers_vcpu` fail msg

himdel (approved):
LGTM :)

PR #430, opened 2026-06-09
[AAP-73765] Candlepin auth implementation

D4D3VD4V3:
## Summary

Load CANDLEPIN_CONSUMER_CERT/KEY from conf_setting at gather time, validate expiry, and use mTLS for ingress upload when possible. Fall back to the existing service-account OAuth2 flow when the cert is missing, invalid, or mTLS fails at upload time. Strip PEM material from config.json payloads.

[AAP-73765](https://issues.redhat.com/browse/AAP-73765)

## Description

- **What is being changed?**
  - `handle_crc_ship_target()` reads Candlepin identity cert/key PEMs from AAP `conf_setting` (`CANDLEPIN_CONSUMER_CERT`, `CANDLEPIN_CONSUMER_KEY`) and passes them into `billing_provider_params`.
  - `PackageCRC` selects auth mode at ship time: valid cert + key → mTLS upload to CRC ingress; otherwise → existing OAuth2 service-account flow (SSO token → bearer upload).
  - On mTLS `SSLError`, falls back to service-account auth when `METRICS_UTILITY_SERVICE_ACCOUNT_ID` / `METRICS_UTILITY_SERVICE_ACCOUNT_SECRET` are configured.
  - Billing collector strips `candlepin_cert_pem` / `candlepin_key_pem` from `config.json` before packaging (PEMs are auth-only, not payload data).
  - New minimal `metrics_utility/library/candlepin/` helpers: cert validity check + secure temp PEM files for mTLS.
  - Adds `cryptography` dependency and unit tests for DB fetch, ship-target wiring, and `PackageCRC` auth behavior.

- **Why is this change needed?**
  - AAP deployments with a Candlepin consumer identity cert in the DB should authenticate CRC uploads via mTLS instead of requiring service-account credentials on every run.
  - Service-account OAuth2 remains the fallback for environments without a cert, with an expired cert, or when mTLS fails at upload time.

- **How does this change address the issue?**
  - At gather time, metrics-utility checks for a valid Candlepin cert in the DB and prefers mTLS for ingress upload.
  - If mTLS is unavailable or fails, behavior matches the existing CRC flow: OIDC token from SSO → upload to `console.redhat.com/api/ingress/v1/upload`.

Assisted By: Claude Sonnet 4.6


metrics_utility/automation_controller_billing/collector.py
```diff
@@ -94,10 +94,13 @@ def _gather_config(self):
         if not super()._gather_config():
             return False
 
-        # Extend the config collection to contain billing specific info:
+        # Extend the config collection to contain billing specific info.
+        # Strip mTLS key material — only needed at ship time, never in the uploaded payload.
+        _SENSITIVE_BILLING_KEYS = {'candlepin_cert_pem', 'candlepin_key_pem'}
+        safe_billing_provider_params = {k: v for k, v in (self.billing_provider_params or {}).items() if k not in _SENSITIVE_BILLING_KEYS}
```
cshiels-ie, line 100: `_SENSITIVE_BILLING_KEYS` is recreated on every call to `_gather_config()`. Move it to module or class level — it is a genuinely constant value and there is no reason to allocate a new `set` on every invocation.

```python
# module level
_SENSITIVE_BILLING_KEYS = frozenset({'candlepin_cert_pem', 'candlepin_key_pem'})
```

metrics_utility/automation_controller_billing/package/package_crc.py
```diff
@@ -11,21 +11,38 @@
 import metrics_utility.base as base
 
 from metrics_utility.exceptions import FailedToUploadPayload
+from metrics_utility.library.candlepin.client import temp_cert_files
+from metrics_utility.library.candlepin.lifecycle import is_cert_valid as _is_cert_valid
 from metrics_utility.logger import logger
 
 
+__all__ = ['PackageCRC', '_is_cert_valid']
+
```
cshiels-ie, line 20: `__all__` is exporting a name with a leading underscore, which is contradictory — the underscore signals "private/internal", `__all__` signals "public API". Either rename the import alias to drop the underscore:

```python
from metrics_utility.library.candlepin.lifecycle import is_cert_valid
```

or remove `_is_cert_valid` from `__all__` entirely. Tests can still import it directly; `__all__` is only needed for `from package import *` consumers.

metrics_utility/automation_controller_billing/package/package_crc.py
```diff
@@ -108,6 +167,17 @@ def _send_data(self, url, files, session):
                 timeout=(31, 31),
             )
 
+        elif self.shipping_auth_mode() == self.SHIPPING_AUTH_CERTIFICATES:
+            proxies = {'https': self.get_proxy_url()} if self.get_proxy_url() else {}
```
cshiels-ie, line 171: **Potential bug — client cert not passed to `session.post`.**

This branch does not pass `cert=` to `session.post(...)`. For mTLS to work the client certificate must be presented in the request. If the base `Package` class sets `session.cert = self._get_client_certificates()` during session setup (before `_send_data` is called), then the override at line 46–49 would cover it — but that needs explicit verification.

If the base does *not* do that, the temp files are written and cleaned up but no client cert is ever sent to the server — the upload silently falls back to unauthenticated. Please confirm with a comment or test. If the cert is not wired in at session level, the fix is:

```python
response = session.post(
    url,
    files=files,
    cert=self._get_client_certificates(),
    verify=self.CERT_PATH,
    proxies=proxies,
    headers=session.headers,
    timeout=(31, 31),
)
```

metrics_utility/library/candlepin/lifecycle.py
```diff
@@ -0,0 +1,58 @@
...
+    """Return True if cert_pem is parseable, already valid, and not yet expired.
+
+    Logs a warning when the cert is not yet valid, expired, or unparseable, then
+    returns False so the caller can fall back to service-account authentication.
+    """
+    try:
+        info = parse_cert(cert_pem)
+        now = datetime.now(timezone.utc)
+        not_before = datetime.fromisoformat(info['not_before'])
+        if now < not_before:
```
cshiels-ie, line 49: **Python ≤ 3.10 compat issue — unnecessary ISO string round-trip.**

`parse_cert()` already computes `cert.not_valid_before_utc` as a timezone-aware `datetime`. Serialising it to an ISO string and then parsing it back with `datetime.fromisoformat` is fragile: on Python ≤ 3.10, `fromisoformat` does not accept the `+00:00`/`Z` suffix produced by `.isoformat()` on UTC datetimes. The result is a `ValueError` — which is silently caught by the `except` below, causing a valid cert to be treated as unparseable.

Suggestion: add `not_before_dt` / `not_after_dt` keys (as `datetime` objects) to the dict returned by `parse_cert`, and use those directly here. Or fold the validity window check into `parse_cert` itself.

metrics_utility/library/candlepin/lifecycle.py
```diff
@@ -0,0 +1,58 @@
...
+            logger.warning(f'Candlepin cert is not yet valid (not_before={info["not_before"]}); falling back to service account auth')
+            return False
+        if info['days_remaining'] < 0:
+            logger.warning(f'Candlepin cert expired at {info["not_after"]}; falling back to service account auth')
```
cshiels-ie, line 53: Minor readability nit: `timedelta.days` is the integer floor (e.g., a cert that expired 12 hours ago has `.days == -1`, which is correct). But the semantics are non-obvious to a reader unfamiliar with `timedelta`.

Consider using `remaining.total_seconds() < 0` instead — it is unambiguous and does not require knowing that `.days` floors toward negative infinity for sub-day intervals.

metrics_utility/management/validation.py
```diff
@@ -77,6 +78,10 @@
 VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
 VALID_SHIP_TARGET_GATHER = {'directory', 's3', 'crc'}
 
+CANDLEPIN_CERT_SETTING_KEY = 'CANDLEPIN_CONSUMER_CERT'
+CANDLEPIN_KEY_SETTING_KEY = 'CANDLEPIN_CONSUMER_KEY'
+CANDLEPIN_CONF_SETTING_QUERY = 'SELECT key, value FROM conf_setting WHERE key IN (%s, %s)'
+
```
cshiels-ie, line 84: Two issues with this constant:

1. **Hardcoded arity** — the two `%s` placeholders are tightly coupled to the two-element `keys` list. Adding a third setting key requires updating this in two places. Consider generating placeholders dynamically:
   ```python
   # build at call site: ', '.join(['%s'] * len(keys))
   CANDLEPIN_CONF_SETTING_QUERY = 'SELECT key, value FROM conf_setting WHERE key IN ({placeholders})'
   ```

2. **DB adapter style** — `%s` is the psycopg/MySQL style. SQLite uses `?`. The tests mock the cursor so this doesn't fail in CI, but a raw SQL test against a SQLite fixture would break. Either document that this requires psycopg, or use the Django ORM to let the adapter handle quoting:
   ```python
   from django.apps import apps
   ConfSetting = apps.get_model('main', 'ConfSetting')  # adjust app label as needed
   rows = ConfSetting.objects.filter(key__in=keys).values_list('key', 'value')
   ```

metrics_utility/management/validation.py
```diff
@@ -164,6 +169,29 @@ def handle_not_s3():
...
+    """Read Candlepin identity cert and key PEMs from conf_setting.
+
+    Returns:
+        Tuple of ``(cert_pem, key_pem)``, either of which may be None if absent
+        or on any DB error. Best-effort: failures are logged and never propagate.
+    """
+    keys = [CANDLEPIN_CERT_SETTING_KEY, CANDLEPIN_KEY_SETTING_KEY]
+    try:
+        from django.db import connection
+
```
cshiels-ie, line 182: Lazy import inside the function body — `from django.db import connection` is re-evaluated on every call. If the reason is to defer import until Django is configured, add a comment explaining that; otherwise a module-level import is cleaner and more conventional for this codebase.

metrics_utility/library/candlepin/client.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        if path and os.path.exists(path):
+            os.unlink(path)
+    except OSError as e:
+        logger.warning(f'Could not remove temp cert file {path}: {e}')
+
+
+@contextmanager
+def temp_cert_files(cert_pem, key_pem):
+    """Context manager: writes cert + key to secure temp files, cleans up on exit."""
```
cshiels-ie, line 41: `temp_cert_files` does not guard against `None` or empty string inputs. If called with `cert_pem=None`, `_write_temp_pem(None)` writes the literal string `"None"` to the temp file — silent corruption rather than an obvious error.

Add an early guard so the failure is immediate and descriptive:

```python
@contextmanager
def temp_cert_files(cert_pem, key_pem):
    if not cert_pem or not key_pem:
        raise ValueError('cert_pem and key_pem must both be non-empty strings')
    ...
```

metrics_utility/library/candlepin/client.py
```diff
@@ -0,0 +1,52 @@
...
+
+from metrics_utility.logger import logger
+
+
+def _write_temp_pem(content):
+    """Write PEM content to a secure temp file. Returns the file path."""
+    fd, path = tempfile.mkstemp(prefix='candlepin_', suffix='.pem')
+    try:
+        os.chmod(path, 0o600)
+        with os.fdopen(fd, 'w') as f:
```
cshiels-ie, line 16: Minor security note: on Python < 3.12, `tempfile.mkstemp()` creates the file with the process umask applied, so there is a brief window between file creation and `os.chmod(path, 0o600)` where group-read may be permitted (e.g., if umask is `0o022`, the file is briefly `0o644`).

On Python 3.12+ `mkstemp` always creates with mode `0o600`, so this is handled automatically. Worth documenting the minimum Python version if < 3.12 support is intended, or accept this as low-risk in a containerised environment.

cshiels-ie (commented):
Claude Automated inline review — see individual comments for findings and suggested fixes.

Assisted by: Sonnet 4.6 (1M context)

metrics_utility/management/validation.py
```diff
@@ -77,6 +78,10 @@
 VALID_SHIP_TARGET_BUILD = {'directory', 's3', 'controller_db'}
 VALID_SHIP_TARGET_GATHER = {'directory', 's3', 'crc'}
 
+CANDLEPIN_CERT_SETTING_KEY = 'CANDLEPIN_CONSUMER_CERT'
+CANDLEPIN_KEY_SETTING_KEY = 'CANDLEPIN_CONSUMER_KEY'
```
cshiels-ie, line 82: **Blocker — key names do not match what AWX actually writes.**

AWX PR [ansible/awx#16388](https://github.com/ansible/awx/pull/16388) (now merged) stores the Candlepin cert under these `conf_setting` keys:

```python
'CANDLEPIN_CERT_PEM'       # this PR reads 'CANDLEPIN_CONSUMER_CERT' ❌
'CANDLEPIN_KEY_PEM'        # this PR reads 'CANDLEPIN_CONSUMER_KEY'  ❌
'CANDLEPIN_CONSUMER_UUID'  # not read here, but available
```

With the current constants, `cursor.fetchall()` returns zero rows on every run. The function silently returns `(None, None)` and logs "No Candlepin identity cert found in DB" — mTLS is never used.

**Fix:**
```python
CANDLEPIN_CERT_SETTING_KEY = 'CANDLEPIN_CERT_PEM'
CANDLEPIN_KEY_SETTING_KEY  = 'CANDLEPIN_KEY_PEM'
```

metrics_utility/automation_controller_billing/package/package_crc.py
cshiels-ie, line 171: **Blocker — mTLS upload must target `cert.console.redhat.com`, not `console.redhat.com`.**

The Jira AC (AAP-73765) requires:
> Analytics tarball uploads use mTLS with the Candlepin certificate to **`cert.console.redhat.com/api/ingress/v1/upload`**

`cert.console.redhat.com` is a distinct CRC ingress endpoint that negotiates client certificates. Posting to the standard `console.redhat.com` endpoint (used for service-account auth) will not accept a client certificate.

This branch uses `url` as passed from `_send_data`'s caller, which comes from `get_ingress_url()` — the same URL used for service-account auth.

**Fix:** add a dedicated method and env var for the cert ingress URL:
```python
def get_cert_ingress_url(self):
    return os.environ.get(
        'METRICS_UTILITY_CRC_CERT_INGRESS_URL',
        'https://cert.console.redhat.com/api/ingress/v1/upload',
    )
```
Use `get_cert_ingress_url()` as the URL in this branch and validate it in `is_shipping_configured()`.

---

**Side note:** the spike's `_send_data` had the comment `# session.cert is already set by base ship()` explaining why `cert=` is not passed explicitly to `session.post`. That comment is missing here — worth restoring so future readers don't have to trace through the base class to understand the contract.

metrics_utility/automation_controller_billing/package/package_crc.py
```diff
@@ -48,19 +65,61 @@ def _get_rh_password(self):
...
     def is_shipping_configured(self):
-        # TODO: move to base, or children
-        ret = super()
+        ret = super().is_shipping_configured()
         if ret is False:
             return False
 
+        if self.shipping_auth_mode() == self.SHIPPING_AUTH_CERTIFICATES:
+            if not self.get_ingress_url():
+                logger.error('METRICS_UTILITY_CRC_INGRESS_URL is not set')
```
cshiels-ie, line 119: Once the cert ingress URL is separated (see comment on `_send_data`), this check should call `get_cert_ingress_url()` rather than `get_ingress_url()` — otherwise a misconfigured cert URL would pass validation but fail at upload time.

metrics_utility/management/validation.py
```diff
@@ -196,6 +224,14 @@ def handle_crc_ship_target():
         allowed = '", "'.join(['controller_db', 'directory', 's3'])
         logger.warning(f'Ignoring METRICS_UTILITY_SHIP_PATH used without METRICS_UTILITY_SHIP_TARGET="{allowed}"')
 
+    cert_pem, key_pem = _fetch_candlepin_cert_from_db()
+    if cert_pem and key_pem:
+        billing_provider_params['candlepin_cert_pem'] = cert_pem
+        billing_provider_params['candlepin_key_pem'] = key_pem
+        logger.info('Candlepin identity cert loaded from DB; mTLS will be attempted.')
+    else:
```
cshiels-ie, line 232: Minor: with the correct key names in place, it would be worth logging a near-expiry warning here so operators know the cert is approaching expiry before the Controller renewal kicks in:

```python
if cert_pem and key_pem:
    from metrics_utility.library.candlepin.lifecycle import parse_cert
    try:
        info = parse_cert(cert_pem)
        if info['days_remaining'] < 30:
            logger.warning(
                f'Candlepin cert expires in {info["days_remaining"]} day(s); '
                'ensure Controller cert renewal is running.'
            )
    except Exception:
        pass
    billing_provider_params['candlepin_cert_pem'] = cert_pem
    ...
```

Not a blocker, but aligns with "Handle certificate validity checks and expiry scenarios gracefully" in the AC.

cshiels-ie (commented):
## Spec alignment review — AAP-73765 + AWX PR #16388

Reviewed against [AAP-73765](https://redhat.atlassian.net/browse/AAP-73765) and the upstream Controller implementation in [ansible/awx#16388](https://github.com/ansible/awx/pull/16388) (merged). Two blockers before this is functional:

### 1. Wrong DB key names — will silently fail on every run (blocker)

The AWX Controller stores the Candlepin cert in `conf_setting` under these keys (confirmed from awx#16388):

| AWX writes | This PR reads | Match? |
|---|---|---|
| `CANDLEPIN_CERT_PEM` | `CANDLEPIN_CONSUMER_CERT` | ❌ |
| `CANDLEPIN_KEY_PEM` | `CANDLEPIN_CONSUMER_KEY` | ❌ |
| `CANDLEPIN_CONSUMER_UUID` | _(not read)_ | — |

As written, `_fetch_candlepin_cert_from_db()` will always return `(None, None)`. Every run will log "No Candlepin identity cert found in DB" and fall back to service-account auth — mTLS will never be used.

### 2. mTLS must target `cert.console.redhat.com`, not `console.redhat.com` (blocker)

The Jira AC says uploads must go to `cert.console.redhat.com/api/ingress/v1/upload`. That is a different endpoint from the service-account path (`console.redhat.com`). The current code sends the mTLS upload to whatever `METRICS_UTILITY_CRC_INGRESS_URL` is set to — typically the service-account URL — which will not accept a client certificate.

### 3. Scope reductions vs spike (low — confirm)

The spike included consumer UUID tracking, near-expiry warnings, check-in, and proactive renewal. None are in this PR. That appears intentional per the Jira (Controller handles lifecycle), but near-expiry logging would be easy to add and improve operator visibility.

See inline comments for exact line references.

metrics_utility/management/validation.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/automation_controller_billing/package/package_crc.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/automation_controller_billing/package/package_crc.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/automation_controller_billing/package/package_crc.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/library/candlepin/lifecycle.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/library/candlepin/lifecycle.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/library/candlepin/client.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/management/validation.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/management/validation.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/automation_controller_billing/collector.py
D4D3VD4V3: Fixed, thanks!

metrics_utility/management/validation.py
D4D3VD4V3: Fixed 1; Documented for 2.

metrics_utility/library/candlepin/client.py
D4D3VD4V3: Documented, thanks!

PR #431, opened 2026-06-10
AAP-75305 - chore: upgrade pandas to 3.0 and openpyxl to 3.1.5 with compat fixes

cshiels-ie:
## Summary

Upgrades pandas from 2.x to 3.0.3 and openpyxl from 3.1.2 to 3.1.5, and fixes all breaking changes introduced by pandas 3.0.

### Version bumps
- `pyproject.toml`: `openpyxl==3.1.2` → `~=3.1.5`, `pandas>=2.2.3` → `~=3.0` (caps at <4.0.0)
- `setup.cfg`: `pandas~=2.2.1` → `~=3.0`, `openpyxl~=3.1.2` → `~=3.1.5`
- `uv.lock`: resolved openpyxl 3.1.5, pandas 3.0.3

### Breaking changes fixed

**1. NaN passed as `float` in string columns** (`dataframe_content_usage.py`, `main_jobevent.py`)

pandas 3.0 passes `float('nan')` instead of `None` for missing values in string columns. `re.match(pattern, float('nan'))` raises `TypeError`. Fixed `extract_collection_name()` and `extract_role_name()` to guard with `isinstance(x, str)` instead of `x is None`.

**2. NaN in `str.join()` filter** (`renewal_guidance.py`)

`v is not None` did not filter `float('nan')`. Replaced with `isinstance(v, str)`.

**3. NaN in `serials` cell causes TypeError in dedup** (`ccsp.py`)

pandas 3.0 stores missing values as `float('nan')` in object-dtype columns. `bool(float('nan'))` is `True` in Python, so the bare `if serials:` guard passed for NaN — causing `for serial in serials:` to raise `TypeError` on a non-iterable scalar. Fixed by replacing the guard with `isinstance(serials, Iterable) and not isinstance(serials, (str, bytes))` (using `collections.abc.Iterable`), which safely skips NaN, `None`, and bare strings. The inner loop retains its existing `isinstance(serial, str) and serial` check to reject empty-string serials.

**4. Deprecated frequency alias** (`test_csv_utils.py`)

pandas 3.0 removed the `'H'` hour alias. Updated to lowercase `'h'`.

### Follow-up hardening (post-review)

- **Version cap**: `pandas>=3.0.0` → `pandas~=3.0` in both `pyproject.toml` and `setup.cfg` to prevent silent breakage from a future pandas 4.0 release.
- **Import cleanup**: Removed inline `import pandas as _pd` from inside `normalize_for_comparison()` in the CCSPv2 test; uses the existing module-level `import pandas` instead.
- **Regression tests**: Added `test_df_to_mapping_nan_float_serials` and `test_df_to_mapping_empty_string_serials` to `TestDedupCCSP` to directly cover the pandas 3.0 dedup bucketing bug.

## Test plan

- [x] `test_dataframe_to_csv_files_large_splitting` — `freq='H'` → `'h'`
- [x] `test_complex_CCSP_sanity_check` — `float` NaN / `re.match` fix
- [x] `test_renewal_guidance` — `join()` filter fix
- [x] `test_big_test_job*` — anonymized rollups suite
- [x] `test_empty_data_for_CCSP_and_CCSPv2` — no regressions
- [x] All existing Candlepin tests — no regressions from ccsp.py change
- [x] `test_df_to_mapping_nan_float_serials` — pandas 3.0 NaN dedup regression
- [x] `test_df_to_mapping_empty_string_serials` — empty serial filtering

## AI Assistance
This change was developed with assistance from Claude Code (Claude Sonnet 4.6).


setup.cfg
```diff
@@ -11,8 +11,8 @@ zip_safe = False
 include_package_data = False
 
 install_requires =
-    pandas~=2.2.1
-    openpyxl~=3.1.2
+    pandas~=3.0
```
hunterkepley, line 14: Should we be going with the latest `3.Y`? I know python pkgs havent been the vector, but with the supply chain attacks lately, i feel we might want to explicitly version when possible

pyproject.toml
```diff
@@ -12,8 +12,8 @@ dependencies = [
     "distro==1.9.0",
     "django==5.2.7",
     "kubernetes>=33.1.0",
-    "openpyxl==3.1.2",
-    "pandas>=2.2.3",
+    "openpyxl==3.1.5",
+    "pandas~=3.0",
```
hunterkepley, line 16: Same question 🙂

pmflanagan (approved):
Looks good to me

hunterkepley (commented):
looks good overall, new tests to accommodate the pandas changes in `dataframe_content_usage.py`, changes to existing tests to accommodate other pandas changes

I had a couple of nitpicks about the test code, but i assume claude was used for that, and we dont care too much, as long as it does the same thing and passes (im still learning to not be nitpicky with AI code)

However left a concern in a comment around versioning

setup.cfg
hunterkepley: I also see that you changed a `~=` to `=` for another pkg, is what sparked the question idea

Pandas latest X stream is 3 so this would be latest version overall i believe, since we dont specify a Z stream here (i think thats how `~=` works? latest version but not past the streams defined)

cshiels-ie: Yeah probaly should leave it with the ~= and not pin openpyxl

hunterkepley (approved):

metrics_utility/automation_controller_billing/dedup/ccsp.py
```diff
@@ -85,9 +86,9 @@ def df_to_mapping(self, df):
             host = row['host_name']
             serials = row['serials']
 
-            if serials:
+            if isinstance(serials, Iterable) and not isinstance(serials, (str, bytes)):
```
himdel: str/bytes is wrong, Iterable is probably fine, it's a set/list or None

metrics_utility/library/dataframes/main_jobevent.py
```diff
@@ -80,7 +80,7 @@ def extract_collection_name(x):
         Returns:
             Two-part collection name string, or None.
         """
-        if x is None:
+        if not isinstance(x, str):
```
himdel, line 83: what's the reason for these,

is this some attempt to introduce typing?

I'd say "x is None" is more specific than checking for strings..

himdel: oh, or would it have to be `x is None or pandas.isNan(x)` now?
I guess it makes sense then

himdel:
Also closes AAP-75306 :+1: 

Seems like the test failures from https://github.com/ansible/metrics-utility/pull/256 are gone, fixed by the `metrics_utility/test/ccspv_reports/dedup/test_complex_CCSPv2_with_canonical_facts/test_complex_CCSPv2_with_canonical_facts.py` changes :tada:

himdel:
looks like the answer to why the incompatible changes in https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#whatsnew-300-prior-deprecations didn't bite us here boils down to:

**1. Copy-on-Write enforced** — The codebase doesn't use chained assignment (`df["a"]["b"] = val`). Mutations go through `.loc[]` or `inplace=True`, both of which work fine under CoW.

**2. String dtype default (`str` instead of `object`)** — The codebase doesn't introspect dtypes or do `dtype == object` checks. String columns are consumed as-is (comparisons, fillna, groupby keys), which all work the same under the new `str` dtype.

**3. Datetime resolution (nanosecond → microsecond)** — All `pd.to_datetime()` calls use `format='ISO8601'` and immediately strip timezone with `.dt.tz_localize(None)`. Billing timestamps don't need nanosecond precision, so the resolution change is invisible. Test assertions use `Timestamp` from strings, which compare equal at either resolution.

**4. `zoneinfo` replaces `pytz`** - The codebase doesn't handle timezone objects at all — it strips them immediately after parsing. No pytz-specific exception handling anywhere. - #152 

**5. Removed deprecated APIs** (`.append()`, `.iteritems()`, `.swaplevel()`, frequency aliases like `M`→`ME`) — None of these are used. The codebase was already on the modern API surface.

**6. `concat()` sorting change** — Every `pd.concat()` call passes `ignore_index=True`, so the axis sorting behavior change for `DatetimeIndex` is irrelevant.

**7. `pd.offsets.Day` semantics** — Not used anywhere.

**8. `inplace=True` now returns `self` instead of `None`** — There are 4 `inplace=True` calls (2x `rename`, 1x `set_index`, 1x `drop`), but none capture the return value, so the change from `None` to `self` has no effect.

(`inplace=True` is newly deprecated for `Resampler.interpolate()` in 3.0, but the 4 existing uses work now)

Also closes #422

himdel (approved):

PR #432, opened 2026-06-10
add mock Prometheus server for vCPU collector integration tests

himdel:
Issue: [AAP-78452](https://redhat.atlassian.net/browse/AAP-78452)

Lightweight Go HTTP server (following the mock-segment-server pattern) that
implements `/api/v1/query` and `/api/v1/query_range` with a configurable CPU
value. Includes `/requests`, `/reset`, and `/config` utility endpoints for
test inspection.

Also adds integration tests that exercise `PrometheusClient` and the
`get_total_workers_cpu`/`get_cpu_timeline` helpers against the real HTTP
server (auto-skipped when the mock isn't running).

Adds a full gather pipeline integration test (`test_vcpu_gather_integration`)
that runs `run_gather_int` with the vCPU collector against mock Prometheus
(metering enabled, JHS disabled), then validates the output tarball against
the `total_workers_vcpu-1.0` JSON schema.

---

Also updates the segment mock to follow the same patterns, most of those come from comments in #409

### DIFF
--- a/metrics_utility/test/gather/test_vcpu_gather_integration.py
+++ b/metrics_utility/test/gather/test_vcpu_gather_integration.py
@@ -1,0 +1,140 @@
+"""Integration test: vCPU gather against mock Prometheus with schema validation."""
+
+import glob
+import io
+import json
+import os
+import tarfile
+import urllib.request
+
+from importlib.resources import files
+from unittest.mock import patch
+
+import jsonschema
+import pytest
+
+from metrics_utility.test.util import run_gather_int
+
+
+MOCK_PROMETHEUS_URL = os.getenv('MOCK_PROMETHEUS_URL', 'http://localhost:9090')
+
+K8S_TOKEN_PATH = '/var/run/secrets/kubernetes.io/serviceaccount/token'
+K8S_CERT_PATH = '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt'
+
+_original_exists = os.path.exists
+_original_open = open
+
+
+def _fake_exists(path):
+    if path in (K8S_TOKEN_PATH, K8S_CERT_PATH):
+        return True
+    return _original_exists(path)
+
+
+def _fake_open(path, *args, **kwargs):
+    if str(path) == K8S_TOKEN_PATH:
+        return io.StringIO('fake-token')
+    return _original_open(path, *args, **kwargs)
+
+
+env_vars = {
+    'METRICS_UTILITY_CLUSTER_NAME': 'test-cluster',
+    'METRICS_UTILITY_DISABLE_JOB_HOST_SUMMARY_COLLECTOR': 'true',
+    'METRICS_UTILITY_DISABLE_SAVE_LAST_GATHERED_ENTRIES': 'true',
+    'METRICS_UTILITY_OPTIONAL_COLLECTORS': 'total_workers_vcpu',
+    'METRICS_UTILITY_PROMETHEUS_URL': MOCK_PROMETHEUS_URL,
+    'METRICS_UTILITY_REPORT_TYPE': 'CCSPv2',
+    'METRICS_UTILITY_SHIP_PATH': './metrics_utility/test/test_data',
+    'METRICS_UTILITY_SHIP_TARGET': 'directory',
+    'METRICS_UTILITY_USAGE_BASED_METERING_ENABLED': 'true',
+}
+
+year = '2025'
+uuid = '00000000-0000-0000-0000-000000000000'
+file_glob = f'./metrics_utility/test/test_data/data/{year}/*/*/{uuid}-*.tar.gz'
+
+
+def _load_schema(filename):
+    return json.loads(files('metrics_utility.schemas').joinpath(filename).read_text())
+
+
+def _configure_mock(**kwargs):
+    data = json.dumps(kwargs).encode()
+    req = urllib.request.Request(f'{MOCK_PROMETHEUS_URL}/config', data=data, method='POST')
+    req.add_header('Content-Type', 'application/json')
+    urllib.request.urlopen(req, timeout=5)
+
+
+def _reset_mock():
+    req = urllib.request.Request(f'{MOCK_PROMETHEUS_URL}/reset', method='POST')
+    urllib.request.urlopen(req, timeout=5)
+
+
+@pytest.fixture
+def cleanup_glob():
+    for f in glob.glob(file_glob):
+        os.remove(f)
+    yield
+    for f in glob.glob(file_glob):
+        os.remove(f)
+
+
+@pytest.mark.filterwarnings('ignore::ResourceWarning')
+@patch('metrics_utility.automation_controller_billing.collectors.open', create=True, side_effect=_fake_open)
+@patch('metrics_utility.automation_controller_billing.collectors.os.path.exists', side_effect=_fake_exists)
+def test_vcpu_gather_validates_against_schema(mock_exists, mock_open, cleanup_glob):
+    _reset_mock()
+    _configure_mock(cpu_value='16', empty_result=False)
+
+    run_gather_int(
+        env_vars,
+        {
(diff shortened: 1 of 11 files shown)


metrics_utility/test/library/test_collectors_prometheus_client_integration.py
```diff
@@ -0,0 +1,163 @@
...
+def get_captured_requests():
+    with urllib.request.urlopen(f'{MOCK_PROMETHEUS_URL}/requests', timeout=5) as resp:
+        return json.loads(resp.read())
+
+
+class TestPrometheusClientIntegration:
+    def setup_method(self):
+        reset_mock()
+        configure_mock(cpu_value='16', empty_result=False)
+
```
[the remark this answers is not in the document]
himdel: No, we want the tests to fail if the mock servers don't exist

tools/docker/docker-compose.yaml
```diff
@@ -71,6 +71,10 @@ services:
         condition: service_completed_successfully
       postgres-wait:
         condition: service_completed_successfully
+      mock-segment:
+        condition: service_started
+      mock-prometheus:
+        condition: service_started
```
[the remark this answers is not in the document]
himdel: Good enough, these start faster than pytest

metrics_utility/test/gather/test_vcpu_gather_integration.py
```diff
@@ -0,0 +1,140 @@
...
+            except KeyError:
+                continue
+            if vcpu_member is None:
+                continue
+            vcpu_data = json.load(vcpu_member)
+
+            manifest_member = tar.extractfile('./manifest.json')
+            if manifest_member is not None:
+                manifest_data = json.load(manifest_member)
+            break
```
hunterkepley, line 116: `total_workers_vcpu.json` is protected from a KeyError, but `manifest.json` is not protected

im assuming this is because we always expect a `manifest.json`, but do not always expect a `total_workers_vcpu.json` - but, since this is mock data, maybe its best to have checks for both, or no checks at all, WDYT?

metrics_utility/test/gather/test_vcpu_gather_integration.py
```diff
@@ -0,0 +1,140 @@
...
+    manifest_schema = _load_schema('manifest.jsonschema')
+    jsonschema.validate(instance=manifest_data, schema=manifest_schema)
+    assert 'total_workers_vcpu.json' in manifest_data
+
+    # verify Prometheus was actually called (instant + range query)
+    with urllib.request.urlopen(f'{MOCK_PROMETHEUS_URL}/requests', timeout=5) as resp:
+        captured = json.loads(resp.read())
+    paths = [r['path'] for r in captured]
+    assert '/api/v1/query' in paths
+    assert '/api/v1/query_range' in paths
```
hunterkepley, line 140: [besides that concern](https://github.com/ansible/metrics-utility/pull/432#discussion_r3461626041) this test file looks great

metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py
```diff
@@ -977,7 +977,7 @@ def _validate_all_data(json_data, statistics):
     print('✅ All data value assertions passed!')
 
 
-MOCK_SEGMENT_URL = 'http://localhost:8765'
+MOCK_SEGMENT_URL = os.getenv('MOCK_SEGMENT_URL', 'http://localhost:8765')
```
hunterkepley, line 906: good change with moving the mock segment to an env var, makes a lot of sense for our mock endpoints to have adjustable ports/IPs

metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py
```diff
@@ -1064,7 +1064,8 @@ def test_from_gather_to_json(cleanup_glob):
     print(f'\nSending rollup to mock Segment server at {MOCK_SEGMENT_URL} ...')
 
     # Clear any requests captured from previous test runs.
-    urllib.request.urlopen(f'{MOCK_SEGMENT_URL}/reset', timeout=10)
+    req = urllib.request.Request(f'{MOCK_SEGMENT_URL}/reset', method='POST')
+    urllib.request.urlopen(req, timeout=10)
```
hunterkepley: for prometheus i see we have 5 min timeouts, do we want to keep it as 5? (if so, do we want to keep this as 10?) I see from your reply to coderabbit that these will start faster than pytest, so maybe its not even a concern at all

but i guess its not a concern at all in general, if it starts fast, the timeouts are not important

tools/mock-prometheus-server/main.go
```diff
@@ -0,0 +1,326 @@
...
+	"strconv"
+	"sync"
+	"time"
+)
+
+type record struct {
+	Timestamp string            `json:"timestamp"`
+	Path      string            `json:"path"`
+	Params    map[string]string `json:"params"`
+}
```
hunterkepley, line 41: Go, my beloved...
(i miss inline json mapping 😢 )

tools/mock-segment-server/main.go
```diff
@@ -1,22 +1,20 @@
 // Mock Segment HTTP server for integration tests.
 //
-// Accepts any POST request, returns {"success":true} with HTTP 200, and appends
-// each request body as a JSONL line (with a UTC timestamp) to an output file.
+// Accepts any POST request, returns {"success":true} with HTTP 200, and
+// captures each request body in memory.
```
hunterkepley, line 4: ahhh, now the removal of that temp directory in the docker-compose file makes sense

hunterkepley (commented):
looks pretty good overall, some comments, but nothing really that important

metrics_utility/test/gather/test_vcpu_gather_integration.py
himdel: Yeah, `manifest.json` not being there is an error, no need to catch the exception, while vcpu would be present only in some tarballs.

metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py
himdel: yeah, it was a review comment in the original segment pr, so I moved it here for consistency :)
plus now we have a test actually using it :)

metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py
himdel: oh good catch, we probably just shouldn't set a timeout at all, there's no telling how fast CI servers would be responding :)

tools/mock-segment-server/main.go
himdel: yeah, the original plan was the server would just dump files,
but with the pattern of using actual requests to pull the data, it's not needed to have 2 ways  :)

hunterkepley (approved):

hunterkepley:
woops, accidentally approved, need to review still 😆

hunterkepley (approved):
looks good, some of the error msgs returned from the mock server are vague (invalid float conversion results in `invalid "<var_name>"` error msg) but since its a mock server it's not important to me

hunterkepley:
thanks for putting this together 🙂

himdel:
Ah good point, updated to add reasons to couple of the error msgs at least :)

```
     === invalid start ===
     invalid start: strconv.ParseFloat: parsing "abc": invalid syntax

     === invalid end ===
     invalid end: strconv.ParseFloat: parsing "xyz": invalid syntax

     === range too large ===
     range too large: 100000000 points (max 10000)                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                               
     === invalid JSON config ===
     invalid JSON: invalid character 'o' in literal null (expecting 'u')
```

hunterkepley (approved):

PR #447, opened 2026-06-18
validator - fix schema path after move to metrics_utility package

himdel:
PR #444 moved schemas from `schemas/` to `metrics_utility/schemas/` but didn't update the path in `tools/validator/validate.py`, so the validator can't find any schemas - updating too.

### DIFF
--- a/tools/validator/validate.py
+++ b/tools/validator/validate.py
@@ -15,7 +15,7 @@ from pathlib import Path
 import jsonschema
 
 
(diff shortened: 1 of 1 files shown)


cshiels-ie (approved):
Separate issue the validator should also need approval from Cloud Analytics

himdel:
> Separate issue the validator should also need approval from Cloud Analytics

Maybe .. I wouldn't want to make it a part of the contract unless they're also running it to check.
Right now, only the schemas are covered, right?

PR #450, opened 2026-06-22
cleanup_glob - fix outlier test fixtures, missing cleanup

himdel:
Follows #448.

Makes more test file cleanup follow the `cleanup_glob` pattern, and fixes broken `cleanup_test_data` fixtures where after-cleanup was commented out.

cleanup_glob:
* `test_from_gather_to_json` - replace before-cleanup with after-cleanup (previously commented out), switch to shared `cleanup_glob` helper; zero-pad month/day in output path for consistency
* `test_gather_jobs_events_summaries_service` - replace before+after cleanup with after-only, using the shared helper
* `test_json` - remove no-op placeholder `cleanup_glob` fixture (test doesn't produce files)
* `test_report_splitting` - switch from `shutil.rmtree` to `cleanup_glob`

cleanup_test_data:
* `test_multiple_files` - narrow `rmtree` scope from `./out` to specific date directories (`./out/data/2025/06/13`, `./out/rollups/2025/06/13`)
* `big_test1/test_all`, `big_test1/test_all_no_events` - narrow `rmtree` scope from `./out` to specific date directories (`./out/data/2024/01/15`, `./out/rollups/2024/01/15`)

### DIFF
--- a/metrics_utility/test/gather/test_gather_jobs_events_summaries_service.py
+++ b/metrics_utility/test/gather/test_gather_jobs_events_summaries_service.py
@@ -1,6 +1,5 @@
 import csv
 import glob
-import os
 
 import pytest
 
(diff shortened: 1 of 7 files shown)


MilanPospisil (approved):
LGTM 👍 

Probably no problem with it, but I see potential problem with cleaning only after test, when some functionality will require clean folder to work properly and the test will not clean after for any reason (failure, process kill), test will fail.

himdel:
Yeah, agreed, the next step will be to make the dirs random, so that this won't happen :)

PR #475, opened 2026-07-02
feat(indirect-nodes): convert collector to daily slicing with collection grouping

pmflanagan:
# feat(indirect-nodes): convert collector to daily slicing with collection grouping

## Description

- What is being changed?
  The `main_indirectmanagednodeaudit` collector is converted from a snapshot collector
  (full-table scan, no time filter) to a daily collector filtered by `main_unifiedjob.finished`.
  The anonymized rollup processor is rewritten to group indirect managed nodes by
  Ansible collection name, counting unique hosts per collection.

- Why is this change needed?
  AWX deletes `IndirectManagedNodeAudit` records after a configurable retention period
  (default 7 days). The snapshot approach produces inconsistent results across customers
  with different retention settings and causes double-counting when summing daily totals.
  Filtering by `job.finished` aligns the collector with calendar-day boundaries and
  eliminates these issues.

  - Adds `since`/`until` parameters to the collector and uses `date_where()` to filter
    by `main_unifiedjob.finished`, matching the pattern used by other daily collectors.
  - Changes the registration from `until_slicing` to `daily_slicing` and bumps the
    version to `2.0`.
  - Rewrites the rollup's `prepare()` to parse the `events` JSON column, extract
    Ansible collection names via `extract_collection_name()`, and group by
    `(organization_name, collection_name)` with unique `host_name` counts per group.
    Organization-level grouping is preserved in intermediate data for on-prem accuracy.
  - Rewrites `merge()` to union host name sets across batches.
  - Rewrites `base()` to strip all PII before the anonymized output leaves the
    customer's environment: `host_names` are removed, and `organization_name` is
    stripped by collapsing org-level groups into collection-level totals with
    re-deduplication of hosts across orgs.
  - Adds `indirect_nodes_by_collection` to the flattened anonymized report output.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Test update

## Self-Review Checklist

- [x] I have performed a self-review of my code
- [x] I have added relevant comments to complex code sections
- [x] I have updated documentation where needed
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases
- [x] I have tested the changes in my local environment

### Prerequisites

- Python 3.12 with project dependencies installed (`pip install -e .`)

### Steps to Test

1. Run collector unit tests:
   `pytest metrics_utility/test/library/test_collectors_main_indirectmanagednodeaudit.py -v`
2. Run rollup unit tests:
   `pytest metrics_utility/test/test_anonymized_rollups/test_indirect_managed_nodes_anonymized_rollup.py -v`
3. Run integration test (requires running AWX database):
   `pytest metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py -v`

### Expected Results

- All tests pass

### DIFF
--- a/metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
+++ b/metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
@@ -2,6 +2,8 @@
 
 import json
 
+import pandas as pd
+
 from metrics_utility.anonymized_rollups.base_anonymized_rollup import BaseAnonymizedRollup
 from metrics_utility.anonymized_rollups.helpers import sanitize_json
 from metrics_utility.automation_controller_billing.dataframe_engine.dataframe_content_usage import (
@@ -81,7 +83,7 @@ class IndirectManagedNodesAnonymizedRollup(BaseAnonymizedRollup):
 
         for _, row in dataframe.iterrows():
             host_name = row.get('host_name')
-            if host_name is None or (isinstance(host_name, float) and host_name != host_name):
+            if pd.isna(host_name):
                 continue
 
             host_name = str(host_name)
--- a/metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
+++ b/metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
@@ -52,6 +52,17 @@ def _make_group_key(organization_name, collection_name):
     return f'{organization_name}{GROUP_KEY_SEPARATOR}{collection_name}'
 
 
+def _aggregate_by_key(groups, key_field):
+    """Aggregate host names across groups by a single field (org or collection)."""
+    aggregated = {}
+    for group in groups.values():
+        key = group[key_field]
+        if key not in aggregated:
+            aggregated[key] = set()
+        aggregated[key].update(group.get('host_names', []))
+    return {k: {'host_names': sorted(v), 'host_count': len(v)} for k, v in sorted(aggregated.items())}
+
+
 class IndirectManagedNodesAnonymizedRollup(BaseAnonymizedRollup):
     """Rollup processor for main_indirectmanagednodeaudit collector data.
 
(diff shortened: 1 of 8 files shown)


metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -164,14 +164,14 @@ def cli_main_host_daily(since, until, output):
     return output.as_files(collector)
 
 
-@register('main_indirectmanagednodeaudit', '1.0', format='csv', fnc_slicing=until_slicing)
+@register('main_indirectmanagednodeaudit', '2.0', format='csv', fnc_slicing=daily_slicing)
```
MilanPospisil: Why change here? This is old collector used in CCSP, please do not change it.

You should ue the new one in library and I see it is there.

Was this LLM mistake or intention for some reason?

metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
```diff
@@ -1,46 +1,126 @@
...
+
+    Groups indirect managed nodes by organization and Ansible collection,
+    counting unique host names per group.
+    """
 
     def __init__(self):
         super().__init__('indirect_managed_nodes')
         self.collector_names = ['main_indirectmanagednodeaudit']
 
     def prepare(self, dataframe):
```
MilanPospisil, line 77: I like it.

Can we also maybe add grouping by organizations only? Because I did not noticed before that one indirect host can have list of collections, not only one, thus one host can appear in more groups (by org and colletion), so grouping by purely organizations and maybe also purely by collections can be nice addition :-) So its clearly visible in data. Sure, we can deduce it from the data, but this way it will be visible on first look.

metrics_utility/automation_controller_billing/collectors.py
pmflanagan: This was a mistake by claude when doing a final check for docstrings and tests and should not have been included, apologies. reverted the change now.

MilanPospisil (approved):
LGTM 👍

PR #480, opened 2026-07-15
Events reshape rollup2

MilanPospisil:
[AAP-69223]

What was changed - the previous events couting was wrong. The final outcome of task is ok, failed, or unreachable event, retries have separate retry event, so we dont have to do complicated logic for detecting it.

The old schema is unchanged besides events of course and we removed total number of hosts that were duplicated variable and always 0 (there is another host variable that counts correctly).

Now we are couting events directly - raw events counts, this is first simple versions. Other counts of more granular level will come next release.

Test:

tools/anonymized_tests/run.py

Should produce some data in out directory. Data will look like this (only event part for modules):

 {
            "collection": "ansible.builtin",
            "collection_source": "certified",
            "jobs_total": 6,
            "jobs_successful_total": 5,
            "jobs_failed_total": 1,
            "tasks_total": 20,
            "jobs_duration_total_seconds": 600.0,
            "jobs_waiting_time_total_seconds": 110.0,
            "jobs_never_started_total": 1,
            "jobs_successful_duration_total_seconds": 510.0,
            "jobs_failed_duration_total_seconds": 90.0,
            "warnings_total": 1,
            "deprecations_total": 1,
            "collected_events_total": 29,
            "event_data_size_total": 2050,
            "ansible_versions": [
                "2.9.10"
            ],
            "runner_on_ok_total": 19,
            "runner_on_failed_total": 2,
            "runner_on_unreachable_total": 1,
            "runner_on_async_ok_total": 1,
            "runner_on_async_failed_total": 2,
            "runner_item_on_ok_total": 1,
            "runner_item_on_failed_total": 2,
            "runner_retry_total": 1,
            "ignore_errors_total": 3,
            "unique_hosts_total": 2
        },


metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -76,71 +102,70 @@ class EventModulesAnonymizedRollup(BaseAnonymizedRollup):
...
+        'runner_item_on_ok_total',
+        'runner_item_on_failed_total',
+        'runner_item_on_failed_ignored_total',
+        'runner_item_on_unreachable_total',
         'warnings_total',
         'deprecations_total',
-        'processed_events_total',
-        'tasks_total',
-        'unique_hosts_total',
+        'events_processed_total',
```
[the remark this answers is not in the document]
MilanPospisil: This can not happen, prepare has finished jobs, another batch will have another finished jobs.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -20,19 +28,49 @@
...
+_RUNNER_EVENTS = frozenset(
+    [
+        'runner_on_ok',
+        'runner_on_async_ok',
+        'runner_item_on_ok',
+        'runner_on_failed',
+        'runner_on_async_failed',
+        'runner_item_on_failed',
+        'runner_on_unreachable',
+        'runner_item_on_unreachable',
```
himdel: seems like an item can only be ok, failed or skipped
the whole loop uses the same connection, so unreachable only happens on the level above

(meaning `runner_item_on_unreachable` is dead code everywhere)

MilanPospisil: This is true, that was already fixed, but I checkout to old commit to make PR smaller, I was sure I will miss something :-)

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -409,172 +452,162 @@ def _filter_and_select_columns(self, dataframe):
...
+        Per-job metrics (counts, durations) are computed over unique job_ids within
+        each group to avoid inflating counts when a job contributes many events.
+        Event-level metrics sum precomputed flag columns — one increment per event.
         """
-        job_failed = task_summary['job_failed']
 
         def _unique_job_sum(col_name):
-            """Return a lambda that sums col_name over one row per unique job_id."""
-            col = task_summary[col_name]
+            col = dataframe[col_name]
```
cshiels-ie: The closures capture `col = dataframe[col_name]` from the outer scope. For `module_stats` and `collection_stats` this is fine — the groupby runs on the same `dataframe`. For `role_stats` the groupby runs on `role_df` (a filtered subset), so `x` carries indices from `role_df` and `col.loc[...]` looks them up in the full frame. This works because pandas preserves the original integer index across boolean filtering — but silently breaks if anyone calls `.reset_index()` between here and the groupby.

Safer to pass the frame explicitly so the closure is self-contained:

```python
def _unique_job_sum(col_name, frame):
    col = frame[col_name]
    def agg(x):
        return col.loc[x[~x.duplicated()].index].sum()
    return agg
```

Then call `_get_common_aggregation(role_df)` separately when building role aggregations, or pass the frame as a parameter.

cshiels-ie (commented):
**CLAUDE REVIEW**

A few things worth a look before merge — three are in the same file, one is a drift risk across two files. Inline comments below.

**Not flagged yet:**
-  /  closures depend on cross-frame index alignment for the role groupby (safe today, fragile to future index resets)
-  rescans the frame once per playbook — groupby is cleaner and O(n)
- Dead  guard in  (already guaranteed by )
-  duplicated across two files — silent drift risk

Everything CodeRabbit flagged is already addressed. The `runner_item_on_unreachable` removal is confirmed intentional.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -409,172 +452,162 @@ def _filter_and_select_columns(self, dataframe):
...
-        role_stats['unique_hosts_total'] = role_stats['host_ids'].apply(lambda x: len(x) if isinstance(x, set) else 0)
 
         return module_stats, collection_stats, role_stats
 
+    def _compute_unique_metadata(self, dataframe):
+        """Compute unique_modules and modules_per_playbook."""
+        unique_modules = sorted(dataframe['module_name'].dropna().unique())
+
+        modules_per_playbook = {}
+        for playbook in dataframe['playbook'].dropna().unique():
```
cshiels-ie: This rescans the full dataframe once per unique playbook — O(n × m) where m is the number of distinct playbooks. For large event frames with many playbooks this adds up.

A groupby is both faster and simpler:

```python
def _compute_unique_metadata(self, dataframe):
    unique_modules = sorted(dataframe['module_name'].dropna().unique())
    modules_per_playbook = (
        dataframe.dropna(subset=['playbook', 'module_name'])
        .groupby('playbook')['module_name']
        .apply(lambda x: sorted(x.unique()))
        .to_dict()
    )
    return unique_modules, modules_per_playbook
```

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -395,11 +438,11 @@ def _filter_and_select_columns(self, dataframe):
 
         if 'ansible_version' not in dataframe.columns:
             dataframe['ansible_version'] = None
+        if 'event_data_length' not in dataframe.columns:
```
cshiels-ie: Dead guard — `_ensure_event_data_length` runs unconditionally earlier in `prepare()` and always adds the column, so this branch can never be reached. Safe to remove.

```suggestion
```

metrics_utility/library/collectors/controller/main_jobevent_service.py
```diff
@@ -9,20 +9,27 @@
 _DEFAULT_JOB_LIMIT = 1_000
 _DEFAULT_ROW_LIMIT = 200_000
 
-_RELEVANT_EVENTS = [
+_RUNNER_EVENTS = [
```
cshiels-ie: `_RUNNER_EVENTS` is also defined as a `frozenset` in `events_modules_anonymized_rollup.py`. They happen to be identical today, but there is nothing enforcing that — if one is updated (e.g. a new event type added) without updating the other, the collector will fetch events the rollup silently ignores, or vice versa.

Consider importing from a single shared location, or at least adding a comment cross-referencing the two definitions.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -394,8 +425,6 @@ def _filter_and_select_columns(self, dataframe):
 
         columns_to_keep = [
             'job_id',
-            'host_id',
```
AparnaKarve, line 397: Without `host_id` we lose the ability to compute `unique_hosts_total` per module/collection — needed for "collection X was used across Z hosts"

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -405,172 +434,165 @@ def _filter_and_select_columns(self, dataframe):
...
+        module_stats = dataframe.groupby(['module_name', 'collection_name', 'collection_source'], as_index=False, observed=True).agg(
             **common_aggregation
         )
-        module_stats = self._compute_tasks_total(module_stats)
-        module_stats['unique_hosts_total'] = module_stats['host_ids'].apply(lambda x: len(x) if isinstance(x, set) else 0)
 
-        collection_stats = task_summary.groupby(['collection_name', 'collection_source'], as_index=False, observed=True).agg(**common_aggregation)
-        collection_stats = self._compute_tasks_total(collection_stats)
-        collection_stats['unique_hosts_total'] = collection_stats['host_ids'].apply(lambda x: len(x) if isinstance(x, set) else 0)
+        collection_stats = dataframe.groupby(['collection_name', 'collection_source'], as_index=False, observed=True).agg(**common_aggregation)
```
AparnaKarve: It looks like `tasks_total` was removed here (previously computed via `_compute_tasks_total()`). It's the simplest measure of "how much did this collection execute" -- needed for PM charts like "top collections by task count."
`jobs_total` is too coarse (job-level), and summing `runner_*_total` risks double-counting in loops. 

Please keep `tasks_total` as a derived field alongside the new runner counters.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
MilanPospisil: Its now for collection.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
MilanPospisil: Added tasks total.

MilanPospisil: Must first clean the DB data so tests have good task_uuid, its defined badly in SQL testing DB right now, I will push it again.

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -338,48 +383,39 @@ def _compute_job_metrics(self, dataframe):
...
+        dataframe['is_warning'] = dataframe['warnings'].apply(self._parse_and_check_json_array).astype(bool)
+        dataframe['is_deprecation'] = dataframe['deprecations'].apply(self._parse_and_check_json_array).astype(bool)
         return dataframe
 
     def _filter_and_select_columns(self, dataframe):
-        """Filter dataframe for required fields and select only needed columns."""
+        """Filter to rows with required fields and keep only needed columns."""
         dataframe = dataframe[
             dataframe['module_name'].notna()
             & dataframe['host_id'].notna()
```
AparnaKarve, line 421: should `task_uuid` also be filtered for `notna()` here?

cshiels-ie (approved):

PR #487, opened 2026-07-18
AWX schema CI - filter backports-zstd on Python 3.14+

himdel:
Follows #461 - awx schema extract job (AAP-75301)

The cron run failed in https://github.com/ansible/metrics-utility/actions/runs/29235262247/job/86768347053

backports-zstd is in awx requirements now, but can't be used with python 3.14 because it's already in stdlib since 3.14

we use whichever version our pyproject allows, which is currently 3.12, 3.13 or 3.14,
so adding a tweak to filter it out in >=3.14


hunterkepley (approved):
LGTM

PR #462, opened 2026-06-25
AAP-80590 | feat: Jira ticket tagging github action

hunterkepley:
# [AAP-80590] Github Action for linking jira MRs to jira tickets automatically

## Description

- What is being changed?

Already working in metrics service: https://redhat-internal.slack.com/archives/C0854435JEL/p1782400578384939

https://github.com/ansible/metrics-service/pull/307#issuecomment-4800890884

Action which automatically links MRs in jira tickets under the `Git Pull Request` section

The action leaves a comment in both cases, when a ticket is found, and when it is not found, in one of 3 places (1. Title, 2. Description, 3. First commit msg). It edits the comments so that it does not spam the MR, and never fails so it does not mock MRs

- Why is this change needed?

The migration to jira cloud got rid of the automatic linking we used to have

- How does this change address the issue?

It links automatically through a script

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ X New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Development environment change
- [ ] Configuration change

## Self-Review Checklist

- [X] I have performed a self-review of my code
- [ ] I have added relevant comments to complex code sections
- [ ] I have updated documentation where needed
- [X] I have considered the security impact of these changes
- [X] I have considered performance implications
- [X] I have thought about error handling and edge cases
- [ ] I have tested the changes in my local environment

### Steps to Test
1. Merge this, create an MR without a ticket, add the ticket after the "negative case" comment is left. It should update the comment and link the MR to the right jira ticket

### Expected Results

The action leaves a comment in both cases, when a ticket is found, and when it is not found, in one of 3 places (1. Title, 2. Description, 3. First commit msg). It edits the comments so that it does not spam the MR, and never fails so it does not mock MRs

## Additional Context

Uses secrets in this repository for jira API + github API access to act as our current bot

PR #494, opened 2026-07-21
jira-pr-link - switch to Python, append instead of overwrite

himdel:
Related to: AAP-83170, #462, ansible/metrics-service#362
Replaces #491

Port the Python-based jira-pr-link workflow from metrics-service (ansible/metrics-service#347, ansible/metrics-service#362) to replace the bash/curl version that blindly overwrote the Jira field on each PR.

The workflow now reads the existing field value, checks for duplicates, and appends — so multiple PRs referencing the same Jira ticket all get linked. Also handles edge cases: null/empty field, plain string from the old v2 API, human-edited content (bullet lists, inlineCards), and substring false-positives.

After this, the workflow files are identical between metrics-utility and metrics-service.

---

wait until https://github.com/ansible/metrics-service/pull/362 gets merged - review there first, then this just copies it here

### DIFF
--- a/.github/workflows/jira-pr-link.yml
+++ b/.github/workflows/jira-pr-link.yml
@@ -91,21 +91,114 @@ jobs:
             exit 0
           fi
 
-          HTTP_CODE=$(curl -s -o response.json -w "%{http_code}" \
-            --connect-timeout 10 \
-            --max-time 30 \
-            -X PUT \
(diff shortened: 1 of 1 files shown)


hunterkepley (approved):
LGTM, this copies changes over from metrics service, and those were approved and merged

PR #496, opened 2026-07-22
fix(indirect-nodes): filter private collections from anonymized Segment payload

pmflanagan:
fix(indirect-nodes): filter private collections from anonymized Segment payload

## Description

- **What is being changed?** `base()` in `IndirectManagedNodesAnonymizedRollup` now
  filters `by_collection` and `by_module` against the public collections whitelist
  (`collections.json`) before the payload is returned and ultimately sent to Segment.
- **Why is this change needed?** Private/custom Ansible collection names (not present
  in the public whitelist) were flowing through unfiltered into the final anonymized
  report, which is transmitted to Segment.com. This was flagged as a requirement: only
  known public collections should appear in the report.
- **How does this change address the issue?**
  - For `by_collection`: the collection name is checked directly against `collections.json`.
  - For `by_module`: `DataframeContentUsage.extract_collection_name()` deduces the
    `namespace.collection` prefix from the full FQCN, and that prefix is checked against
    `collections.json`. This reuses an existing parser already present in the codebase.
  - The `_no_collection` and `_no_module` sentinels are also excluded as they are not
    valid public collection names.
  - `prepare()` output is untouched, so private names remain available in local DB
    storage as intended.

### Prerequisites

None - unit tests are self-contained.

### Steps to Test

1. Run the indirect nodes test suite:
   `.venv/bin/pytest metrics_utility/test/test_anonymized_rollups/test_indirect_managed_nodes_anonymized_rollup.py -v`

### Expected Results

All 40 tests pass, including 4 new tests covering:
- Private collections excluded from `by_collection`
- Private modules (unknown collection prefix) excluded from `by_module`
- `_no_collection` sentinel excluded from `by_collection`
- `_no_module` sentinel excluded from `by_module`

## Required Actions

No external actions required. No schema changes - this change only removes entries
from existing `by_collection` and `by_module` arrays; the structure is unchanged.

### Code Quality


metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
```diff
@@ -90,6 +91,9 @@ class IndirectManagedNodesAnonymizedRollup(BaseAnonymizedRollup):
     def __init__(self):
         super().__init__('indirect_managed_nodes')
         self.collector_names = ['main_indirectmanagednodeaudit']
+        collections_path = os.path.join(os.path.dirname(__file__), 'collections.json')
+        with open(collections_path, 'r') as f:
+            self.known_collections = json.load(f)
```
hunterkepley: This will be the third usage of this exact code in metrics utility,

`metrics_utility/anonymized_rollups/anonymized_rollups.py` and `metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py`

the `anonymized_rollups.py` file has the best implementation, with a try/catch for the `with open` - if this fails, it will crash

Could we move the code for this in `anonymized_rollups.py` into a function in a common file in the `anonymized_rollups` directory and use that for these 3 locations?

```py
collections_path = os.path.join(os.path.dirname(__file__), 'collections.json')
    try:
        with open(collections_path, 'r') as f:
            return json.load(f)
    except (FileNotFoundError, json.JSONDecodeError):
        return {}
```

Sadly, AI is really good at not reusing code, it loves to do this stuff. And apparently it chose the worse version (without error checking). I think it would be good to reuse rather than retype, as now we have 2 locations which have this code without error checking 😛

hunterkepley (commented):
looks mostly good, one concern

metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
pmflanagan: Thanks for this, adding a follow-up commit now.

metrics_utility/anonymized_rollups/helpers.py
```diff
@@ -13,6 +15,21 @@
     HAS_NUMPY = False
 
 
+def load_known_collections():
```
hunterkepley, line 18: perfect, thank you 😄

metrics_utility/anonymized_rollups/events_modules_anonymized_rollup.py
```diff
@@ -149,10 +148,7 @@ def __init__(self):
 
         self.collector_names = ['main_jobevent_service']
 
-        # Open the JSON file using path relative to this module
-        collections_path = os.path.join(os.path.dirname(__file__), 'collections.json')
-        with open(collections_path, 'r') as f:
-            self.collections = json.load(f)
+        self.collections = load_known_collections()
```
hunterkepley, line 151: and with this, `self.collections` will become just an empty dict instead of crashing (or being `None`, and crashing later) 😄

metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
MilanPospisil: Hunter - at least not critical problem, if it will not load this collections list, the whole task should fail anyway and will be catched by task try-catch, but worth catching it there.

However, your solution will silently pass and loaded file will be empty, which will cause silently drop all modules and collections from statistics, because no collection name will match the empty list.

hunterkepley (approved):
LGTM

metrics_utility/anonymized_rollups/indirect_managed_nodes_anonymized_rollup.py
MilanPospisil: Ok, this seems to be edge case anyway, so both solution are acceptable for me:

1) Hunters solution will silently drop all modules, collections and roles stats
2) Throwing exception will fail the whole daily rollup and we will not send any data at all, if this failure will occur only in indirect hosts collector, that would be better to fail it, but fail daily task is not that good

Lesser evil is probably Hunters solution :-) @pmflanagan

MilanPospisil: For the future, we should have some logging mechanism instead of failing the daily rollup. Failing collectors as whole is good though, but this is daily rollup, better to sent at least jobs and other statistics than nothing at all.

Thanks Hunter for catching it.

pmflanagan: Merging this change for now so we can get the filtering in but will work on a long-term solution.

PR #500, opened 2026-07-23
CI - add mock segment server to pytest-service job

himdel:
Issue: AAP-75303

Counterpart to https://github.com/ansible/metrics-service/pull/366 so that m-s tests usign mock segment don't fail here.

Updates the metrics-service test env on metrics-utility CI.
(dups bits of the pytest job in pytest-service, until we unify both fully)

- Build and start mock-segment in the pytest-service job so metrics-service integration tests pass with local metrics-utility
- Add MOCK_SEGMENT_URL to compose metrics-service env anchor


MilanPospisil (approved):
LGTM 👍

PR #477, opened 2026-07-13
[AAP-78076] test(indirect-nodes): add OperationalError propagation test

pmflanagan:
[AAP-78076]

## Description

Adds a missing negative case from the AAP-78076 test plan: verifying that an `OperationalError`
(e.g. DB connection refused) raised by the `main_indirectmanagednodeaudit` collector is not
caught by `cli_main_indirectmanagednodeaudit` and propagates up to the caller, as distinct from
`ProgrammingError` (missing table) which is intentionally caught and logged as a warning.

### Prerequisites

- Postgres running with seed data imported:
  ```
  cat tools/docker/{roles,latest,functions,conf_setting,main_hostmetric,main_instance,main_jobhostsummary,dab_feature_flags,main_indirectmanagednodeaudit}.sql | psql
  ```

### Steps to Test

```
uv run pytest -s -v metrics_utility/test/test_collectors.py
```

### Expected Results

All 5 tests in `test_collectors.py` pass.

### Code Quality

- [x] Code is properly linted and formatted.
- [x] All tests (existing and new) pass successfully.
- [x] Tests are added or updated as needed.
- [x] Changes are reviewed and approved by at least two team members.
- [x] Documentation is updated where applicable.

PR #465, opened 2026-07-01
[AAP-78076] test(indirect-nodes): add null join and orphaned record tests

pmflanagan:
# [AAP-78076] Add null join and orphaned record tests for indirect node collector

## Description

- Adds two unit tests for error handling edge cases in the
  `main_indirectmanagednodeaudit` collector (Part 4 of indirect node
  collection).
- `test_main_indirectmanagednodeaudit_null_join_references`: asserts that
  records with NULL `job_id`, `inventory_id`, or `organization_id` are
  returned without error. The LEFT JOINs in the collector SQL produce NULL
  for the joined columns; the collector does not filter or raise.
- `test_main_indirectmanagednodeaudit_orphaned_records`: asserts that records
  whose referenced job, inventory, or organization has been deleted are
  returned without error, producing the same NULL join behavior.

## Type of Change

- [x] Test update

## Self-Review Checklist

- [x] I have performed a self-review of my code
- [x] I have added relevant comments to complex code sections
- [x] I have updated documentation where needed
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases
- [x] I have tested the changes in my local environment

### Prerequisites

No DB required - tests mock `_copy_table_pandas`.

### Steps to Test

```
uv run pytest metrics_utility/test/library/test_collectors_main_indirectmanagednodeaudit.py -v
```

### Expected Results

All 5 tests pass.

PR #464, opened 2026-06-30
[AAP-81123] Switch indirect node collector from hourly to daily snapshot

pmflanagan:
[AAP-81123](https://redhat.atlassian.net/browse/AAP-81123)

## Description

Switches `main_indirectmanagednodeaudit` from an hourly time-windowed
collector to a daily full-table snapshot, matching the pattern used by
`execution_environments` and `table_metadata`.

- Collector removes `since`/`until` params and the `WHERE created >=`
  filter; registration changes from `daily_slicing` to `until_slicing`.
- `merge()` simplified to a pass-through (cross-batch deduplication is
  not needed with a single daily snapshot; within-snapshot dedup in
  `prepare()` is unchanged).

### Prerequisites

- Postgres running with seed data imported:
  ```
  cat tools/docker/{roles,latest,functions,conf_setting,main_hostmetric,main_instance,main_jobhostsummary,dab_feature_flags,main_indirectmanagednodeaudit}.sql | psql
  ```

### Steps to Test

1. `uv run pytest -s -v metrics_utility/test/library/test_collectors_main_indirectmanagednodeaudit.py`
2. `uv run pytest -s -v metrics_utility/test/test_anonymized_rollups/test_indirect_managed_nodes_anonymized_rollup.py`
3. `uv run pytest -s -v metrics_utility/test/test_anonymized_rollups/test_from_gather_to_json.py`

### Expected Results

All tests pass.

### Code Quality

- [x] Code is properly linted and formatted.
- [x] All tests (existing and new) pass successfully.
- [x] Tests are added or updated as needed.
- [x] Code includes relevant comments for complex sections.
- [ ] Changes are reviewed and approved by at least two team members.
- [ ] Documentation is updated where applicable.

PR #457, opened 2026-06-24
Feature/indirect managed nodes anonymization

pmflanagan:
## Description

- Adds `IndirectManagedNodesAnonymizedRollup` to `hourly_rollup_processors` in `daily_metrics_rollup.py` so the 24 hourly indirect node collections are merged into the `DailyMetricsSummary` each day.
- Passes `indirect_managed_nodes_rollup` to `anonymize_rollups()` in `daily_anonymize_and_prepare.py` so the merged rollup is included in the anonymized Segment payload.
- Adds an `indirect_managed_nodes` example entry to `collect_hourly_metrics` in `TASK_METADATA`.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Self-Review Checklist

- [x] I have performed a self-review of my code
- [x] I have added relevant comments to complex code sections
- [x] I have updated documentation where needed
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases
- [x] I have tested the changes in my local environment

### Prerequisites

- metrics-utility `feature/indirect-managed-nodes-anonymization` must be merged and released to PyPI before this PR's CI will pass (the import of `IndirectManagedNodesAnonymizedRollup` fails against the current published version).

### Steps to Test

1. After metrics-utility is updated on PyPI: `uv run pytest tests/ -v`
2. Verify no new failures beyond the pre-existing baseline.

### Expected Results

Test suite passes. If `indirect_managed_nodes` hourly collections exist in the database, the daily rollup will include a deduplicated count of unique indirect managed nodes and the Segment payload will contain `indirect_managed_nodes` and `rollup_period_indirect_managed_nodes_total`.

PR #502, opened 2026-07-24
[stable-0.7] feat(AAP-78070): indirect managed nodes daily collection

pmflanagan:
## Summary

Backport of the indirect managed nodes feature to stable-0.7.

Cherry-picks 12 commits from devel (PRs #446, #453, #457, #464, #465, #475, #477, #478, #479, #482, #484, #496) plus one fixup commit.

The fixup commit covers things not in the cherry-picks:
- `salt`/`hash` dead code removal (devel #399 was not cherry-picked; #496 requires salt to already be gone)
- `ansible.builtin` added to `collections.json` (already in devel)
- Test assertion updates for the private-collection filter behaviour change (#496 removes unknown entries rather than renaming them to 'Custom') — affects `test_from_gather_to_json.py`, `test_anonymized_rollups_module.py`, `test_multiple_files.py`, and `big_test1/test_all.py`

Note: ruff `pr-checks` will fail on pre-existing violations in unchanged files — these are caught by a newer ruff version and are independent of this PR.

## Test plan

- [x] `uv run pytest` — 453 passed

### DIFF
--- a/metrics_utility/automation_controller_billing/collectors.py
+++ b/metrics_utility/automation_controller_billing/collectors.py
@@ -138,14 +138,14 @@ def cli_main_host_daily(since, until, output):
     return output.as_files(collector)
 
 
-@register('main_indirectmanagednodeaudit', '1.0', format='csv', fnc_slicing=daily_slicing)
+@register('main_indirectmanagednodeaudit', '1.0', format='csv', fnc_slicing=until_slicing)
 def cli_main_indirectmanagednodeaudit(since, until, output):
     if 'main_indirectmanagednodeaudit' not in get_optional_collectors():
         return None
 
     # the table does not exist in 2.4, may be 2.6+
     try:
-        collector = main_indirectmanagednodeaudit(db=connection, since=since, until=until)
+        collector = main_indirectmanagednodeaudit(db=connection)
         return output.as_files(collector)
     except (ProgrammingError, UndefinedTable) as e:
         logger.warning(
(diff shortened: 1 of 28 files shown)


himdel (approved):
LGTM :+1: 

(pr-checks failures fixed in https://github.com/ansible/metrics-utility/pull/507)

metrics_utility/automation_controller_billing/collectors.py
```diff
@@ -138,14 +138,14 @@ def cli_main_host_daily(since, until, output):
     return output.as_files(collector)
 
 
-@register('main_indirectmanagednodeaudit', '1.0', format='csv', fnc_slicing=daily_slicing)
+@register('main_indirectmanagednodeaudit', '1.0', format='csv', fnc_slicing=until_slicing)
```
himdel: Is this part needed for the backport?

This would change how the CLI collects indirect data .

himdel (commented):
Actually, withdrawing approval for now...

The indirect managed collector change would change the CLI behavior too.

metrics_utility/automation_controller_billing/collectors.py
himdel: => There should be no changes to how the CLI indirect collection behaves.

No changes in a_c_b/collectors at all,
the actual collector in library must still accept and respect since/until when given.
(If it helps splitting into 2 collectors, that's ok too, but probably you just need to make since/until optional?)

himdel (approved):

PR #510, opened 2026-07-28
AAP-75313 | feat: segment size up + calculate header size

hunterkepley:
[AAP-75313]
Dont forget to link back issue to PR.

> [!WARNING]
> **Failure to notify downstream stakeholders (such as the Analytics team) of schema or version changes can result in outages and loss of revenue.**
>
> Please notify stakeholders to disseminate the change correctly, most notably the **Analytics HCC service**.
> A critical ticket to track the change prior to promotion.
> The change should not be merged until dependencies have been updated.

## Description

- What is being changed?

1. Segment max size from 24KB -> 32KB
2. Take the header/entire packet size into account with chunking (right now, we only take the content itself into account. This results in inefficient chunking. Whereas, if we subtract the size of everything except the content from the max message size, we get a more accurate chunking of the data)

- Why is this change needed?

Ticket justification: `we’re arbitrarily limiting chunk contents to 24kB to account for the possibility of 8kB of per-chunk overhead, while the overhead is often much smaller, but, also, allowed to grow larger and could still fail.`

It both increases the limit to 32KB (which was one ask of the ticket, which is the max data size for the Segment SDK), as well as attempts to properly calculate the right # of bytes we can take advantage of, by subtracting the header/envelope size from the max. Then, we use this new max when chunking the actual data so it is properly sized

## Required Actions

- [ ] Blocked by PR/MR: #XXX

### Code Quality

- [X] Code is properly linted and formatted.
- [X] All tests (existing and new) pass successfully.
- [X] Tests are added or updated as needed.
- [X] Code includes relevant comments for complex sections.

## Notes for Reviewers

## AI Assistance

Used Claude to go over CodeRabbit's comments and make changes where seen fit. The core of this was written by myself, as well as the tests were manually updated by myself

### DIFF
--- a/metrics_utility/library/storage/segment.py
+++ b/metrics_utility/library/storage/segment.py
@@ -190,9 +190,10 @@ class StorageSegment:
             'context': {},
         }
         properties = self._build_properties(artifact_name, {}, 0, 0, 0)
+        serializable_meta = {k: v.isoformat() if isinstance(v, datetime.datetime) else v for k, v in segment_meta.items()}
         header = {
             **segment_envelope,
-            **segment_meta,
+            **serializable_meta,
             'properties': properties,
             'anonymousId': anonymous_id,
             'event': event_name,
--- a/metrics_utility/library/storage/segment.py
+++ b/metrics_utility/library/storage/segment.py
@@ -49,6 +49,18 @@ class StorageSegment:
         if not self.write_key:
             logger.info('StorageSegment: write_key not set. Analytics will be disabled.')
 
+    def _build_properties(self, artifact_name, data, chunk_number, total_chunks, chunk_size):
+        return {
+            'artifact_name': artifact_name,
+            'data': data,
+            'upload_timestamp': datetime.datetime.now(tz=datetime.UTC).isoformat(),
+            'chunk_info': {
+                'chunk_number': chunk_number,
+                'total_chunks': total_chunks,
+                'chunk_size': chunk_size,
+            },
+        }
+
     def _calculate_size(self, data):
         """Calculate the size of data in bytes."""
         return len(json.dumps(data).encode('utf-8'))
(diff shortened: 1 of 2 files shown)


metrics_utility/library/storage/segment.py
```diff
@@ -155,7 +166,27 @@ def put(self, artifact_name, *, filename=None, fileobj=None, dict=None, event_na
         # Setting to None restores the SDK default (https://api.segment.io).
         analytics.host = self.host or None
 
-        max_size = self.REGULAR_MESSAGE_LIMIT
+        if not segment_meta:
+            segment_meta = {}
+        message_id = segment_meta.get('message_id')
+
+        header = {
```
himdel: I wonder if we'll need to do something to make sure `header` and the later `segment.track` call still get the same data,
on one hand, it seems potentially brittle if what we pass in changes there and not here,
on the other hand this is all in one file and sounds like a bug like that would probably get auto-flagged these days :).

That said, `segment_meta.get('integrations' & 'context', ...` may be a real bug? These would be coming from segment code, so good to count, but never from our `segment_meta`?

What if we split this into 3 parts for more clarity - what we actually pass to segment directly, segment meta splat as is, and what segment adds itself .. (and then maybe the shape of properties with params for values could be a shared helper?)

(But LGTM overall :+1: , only pulling the nonexistent keys looks like a bug)

hunterkepley: Added a helper func for the shape of the properties and split up the overhead calculation into those 3 parts, should be less brittle now 😄

metrics_utility/library/storage/segment.py
```diff
@@ -170,21 +182,17 @@ def put(self, artifact_name, *, filename=None, fileobj=None, dict=None, event_na
...
-                'artifact_name': artifact_name,
-                'data': {},
-                'upload_timestamp': datetime.datetime.now(tz=datetime.UTC).isoformat(),
-                'chunk_info': {'chunk_number': 0, 'total_chunks': 0, 'chunk_size': 0},
-            },
+            'integrations': {},
+            'context': {},
         }
+        properties = self._build_properties(artifact_name, {}, 0, 0, 0)
+        header = {**segment_envelope, 'properties': properties}
```
himdel:
```suggestion
        header = {**segment_envelope, **segment_meta, 'properties': properties}
```

to count `segment_meta` too

(eh, and to be `--pedantic`, `anonymous_id` and `event` belong side by side with `properties` rather than in the envelope, but that's just a semantic difference, doesn't affect the computation :shrug: :) )

himdel (approved):
LGTM, just missing that `segment_meta` now in the computation

metrics_utility/library/storage/segment.py
hunterkepley: Happy to make that change either way 😄

himdel (approved):

PR #511, opened 2026-07-29
add schema and validation script

ShaiahWren:
[AAP-84017]

  - Define TypedDict classes for the anonymized daily rollup payload sent to Segment
  - Add return type annotation to `flatten_json_report()` and `anonymize_rollups()`
  - Add OpenAPI schema generation script using pydantic `TypeAdapter`
  - Add validation script to check rollup JSON against the TypedDict types (strict mode, extra keys rejected)

  ## Context

  aap-metrics and Analytics teams need a shared reference for what
  fields are expected in the Segment payload. The TypedDict classes
  serve as the single source of truth. The OpenAPI schema is
  generated from the types, not hand-maintained.

 ## How to use

  ```bash

  # Generate OpenAPI schema
  uv run python tools/generate_segment_schema.py
  uv run python tools/generate_segment_schema.py -o schema.yaml

  # Validate rollup data
  uv run python tools/validate_segment_contract.py
  path/to/rollup.json
```

  Files

  - metrics_utility/anonymized_rollups/types.py — TypedDict definitions
  - metrics_utility/anonymized_rollups/anonymized_rollups.py — return type annotations
  - tools/generate_segment_schema.py — OpenAPI schema generation
  - tools/validate_segment_contract.py — data validation
  - pyproject.toml — pydantic, pyyaml dev dependencies

  
Assisted by Claude AI

## Notes for Reviewers

### DIFF
--- a/metrics_utility/anonymized_rollups/anonymized_rollups.py
+++ b/metrics_utility/anonymized_rollups/anonymized_rollups.py
@@ -13,6 +13,7 @@ from metrics_utility.anonymized_rollups.jobhostsummary_anonymized_rollup import 
 from metrics_utility.anonymized_rollups.jobs_anonymized_rollup import JobsAnonymizedRollup
 from metrics_utility.anonymized_rollups.table_metadata_anonymized_rollup import TableMetadataAnonymizedRollup
 from metrics_utility.anonymized_rollups.task_executions_anonymized_rollup import TaskExecutionsAnonymizedRollup
+from metrics_utility.anonymized_rollups.types import AnonymizedPayload
 from metrics_utility.automation_controller_billing.dataframe_engine.dataframe_content_usage import DataframeContentUsage
 
 
@@ -439,7 +440,7 @@ def _as_list(value: Any) -> list[Any]:
     return value if isinstance(value, list) else []
 
 
-def flatten_json_report(data: dict[str, Any]) -> dict[str, Any]:
+def flatten_json_report(data: dict[str, Any]) -> AnonymizedPayload:
     """
     Manually flattens the given nested report into:
       - statistics: object of primitive totals (includes credentials)
@@ -572,7 +573,7 @@ def anonymize_rollups(
     task_executions_rollup=None,
     indirect_managed_nodes_rollup=None,
     **kwargs,
-):
+) -> AnonymizedPayload:
     """
     Combines rollup data, flattens it, and anonymizes sensitive fields.
 
--- a/metrics_utility/anonymized_rollups/types.py
+++ b/metrics_utility/anonymized_rollups/types.py
@@ -1,0 +1,178 @@
+"""TypedDict definitions for the anonymized daily rollup payload sent to Segment."""
+
+from typing import NotRequired, TypedDict
+
+
+class Statistics(TypedDict):
+    rollup_period_execution_environments_total: int | None
+    rollup_period_EE_default_total: int | None
+    rollup_period_EE_custom_total: int | None
+    rollup_period_jobs_total: int
+    rollup_period_jobs_successful: int
+    rollup_period_jobs_failed: int
+    rollup_period_jobs_duration_all_statuses_seconds: int | float
+    rollup_period_jobs_successful_duration_total_seconds: int | float
+    rollup_period_jobs_failed_duration_total_seconds: int | float
+    rollup_period_organizations_total: int
+    rollup_period_forks_total: int
+    rollup_period_templates_total: int
+    rollup_period_inventories_total: int
+    rollup_period_unique_hosts_total: int
+    rollup_period_job_host_pairs_total: int | None
+    rollup_period_successful_hosts_total: int
+    rollup_period_failed_hosts_total: int
+    rollup_period_unreachable_hosts_total: int
+    rollup_period_indirect_managed_nodes_all_total: int
+    rollup_period_tasks_total: int
+    rollup_period_task_ok_total: int
+    rollup_period_task_failed_total: int
+    rollup_period_task_skipped_total: int
+    rollup_period_task_unreachable_total: int
+    rollup_period_task_ignored_total: int
+    # Present only when collected_events_total > 0
+    rollup_period_modules_total: NotRequired[int | None]
+    rollup_period_collected_events_total: NotRequired[int]
+    rollup_period_warnings_total: NotRequired[int]
+    rollup_period_deprecations_total: NotRequired[int]
+    rollup_period_playbooks_total: NotRequired[int]
+
+
+class JobStatsFields(TypedDict):
+    jobs_total: int
+    jobs_failed_total: int
+    jobs_successful_total: int
+    jobs_never_started_total: int
+    jobs_duration_total_seconds: int | float
+    jobs_successful_duration_total_seconds: int | float
+    jobs_failed_duration_total_seconds: int | float
+    job_duration_maximum_seconds: int | float | None
+    job_duration_minimum_seconds: int | float | None
+    job_waiting_time_total_seconds: int | float
+    job_waiting_time_maximum_seconds: int | float | None
+    job_waiting_time_minimum_seconds: int | float | None
+    templates_total: int
+    inventories_total: int
+
+
+class HostSummaryFields(TypedDict):
+    unreachable_total: int
+    failed_total: int
+    ok_total: int
+    skipped_total: int
+    ignored_total: int
+    rescued_total: int
+    successful_hosts_total: int
+    failed_hosts_total: int
+    unreachable_hosts_total: int
+
+
+class JobsByJobType(JobStatsFields, HostSummaryFields):
+    job_type: str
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Review by himdel — CHANGES_REQUESTED** [review]
I haven't thoroughly reviewed the openapi schema itself yet, and this might make sense as an intermediary first step, but:

* the point of openapi over jsonschema is that you can generate openapi schemas from the actual code - after refactoring the code to match what the generators can handle - I think you'll probably need to define a model class, and generate both the schema and the payload through that class, though maybe there's a way to generate it from a well typed generator function too, which might be simpler (EDIT: is simpler, see below :))

* do NOT use `metrics_utility/schemas/` unless the team agrees to it, that location is specifically protected to ping the Analytics team on any changes - we agreed to such a contract for billing, I don't know if we have such a contract for anonymous data - notifying analytics is probably a good idea, but not in any way that could be confused for billing related issues.

EDIT:

Something like this should work in principle...

```python
  # in anonymized_rollups.py (or a nearby types module)
  class ModuleStatsItem(TypedDict):
      module: str
      collection: str
      jobs_total: int
      ...

  class AnonymizedPayload(TypedDict):
      statistics: Statistics
      module_stats: list[ModuleStatsItem]
      ...

  def flatten_json_report(...) -> AnonymizedPayload:                                                                                                                                                                                                               
      ... # pretty much as is, with the added type annotation

  # in a generate_schema.py script
  from pydantic import TypeAdapter
  from metrics_utility.anonymized_rollups.types import AnonymizedPayload

  schema = TypeAdapter(AnonymizedPayload).json_schema()
```

(and then we need to run it, in pre-commit or pr-checks, or both :))
(and there might be a follow-up task to actually enable type-based linters in our code, and fix any surprises)

metrics_utility/anonymized_rollups/anonymized_rollups.py
```diff
@@ -560,18 +561,18 @@ def flatten_json_report(data: dict[str, Any]) -> dict[str, Any]:
...
-    credentials_rollup,
-    table_metadata_rollup,
-    controller_version_rollup,
+    events_modules_rollup: Any,
+    execution_environments_rollup: Any,
+    jobs_rollup: Any,
+    job_host_summary_rollup: Any,
+    credentials_rollup: Any,
+    table_metadata_rollup: Any,
+    controller_version_rollup: Any,
```
AparnaKarve: Nit: these params could reference the actual rollup classes (`EventModulesAnonymizedRollup`, `JobsAnonymizedRollup`, etc.) rather than `Any` -- any reason not to?

metrics_utility/anonymized_rollups/types.py
```diff
@@ -0,0 +1,175 @@
...
+    jobs_successful_duration_total_seconds: float
+    jobs_failed_duration_total_seconds: float
+    job_duration_maximum_seconds: float | None
+    job_duration_minimum_seconds: float | None
+    job_waiting_time_total_seconds: float
+    job_waiting_time_maximum_seconds: float | None
+    job_waiting_time_minimum_seconds: float | None
+    templates_total: int
+    inventories_total: int
+    ansible_versions: list[str]
```
himdel: isn't `ansible_version` specific to only some of these? Looks like `JobsByAnsibleVersion` wouldn't have it?
(no `ansible_versions_aggregation ` in `_aggregate_by_ansible_version`)

metrics_utility/anonymized_rollups/types.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class CollectionStats(EventStatsFields):
+    collection: str
+    unique_hosts_total: int
+
+
+class RoleStats(EventStatsFields):
+    role: str
+    collection: str
```
himdel: `collection: str | None` for bare role support?

metrics_utility/anonymized_rollups/types.py
```diff
@@ -0,0 +1,175 @@
...
+    rollup_period_task_failed_total: int
+    rollup_period_task_skipped_total: int
+    rollup_period_task_unreachable_total: int
+    rollup_period_task_ignored_total: int
+    # Present only when collected_events_total > 0
+    rollup_period_modules_total: NotRequired[int | None]
+    rollup_period_collected_events_total: NotRequired[int | None]
+    rollup_period_warnings_total: NotRequired[int]
+    rollup_period_deprecations_total: NotRequired[int]
+    rollup_period_playbooks_total: NotRequired[int | None]
```
himdel: `len` always returns an int, and the other branch doesn't add the key, isn't this just `NotRequired[int]`?

tools/generate_segment_schema.py
```diff
@@ -0,0 +1,36 @@
...
+from metrics_utility.anonymized_rollups.types import AnonymizedPayload
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Generate JSON Schema for the Segment anonymized payload.')
+    parser.add_argument('-o', '--output', help='Write to file instead of stdout')
+    args = parser.parse_args()
+
+    schema = TypeAdapter(AnonymizedPayload).json_schema()
+    output = json.dumps(schema, indent=2)
```
himdel: probably one more step here...

the json schema this creates is in the right format for openapi, but isn't the openapi schema itself..
so we still need to wrap it in the openapi thingy and output as yaml instead of json...


```suggestion
    schema = TypeAdapter(AnonymizedPayload).json_schema(
        ref_template='#/components/schemas/{model}', # changes "$ref"
    )
    
    defs = schema.pop('$defs', {})
    openapi = {
        'openapi': '3.1.0',
        'info': {'title': 'Segment Anonymized Payload', 'version': '1.0.0'},
        'paths': {},
        'components': {
            'schemas': {
                'AnonymizedPayload': schema,
                **defs,
            },
        },
    }
    
    output = yaml.dump(openapi, default_flow_style=False, sort_keys=False)
```

(+- an import yaml)

tools/validate_segment_contract.py
```diff
@@ -0,0 +1,47 @@
...
+
+    json_path = Path(sys.argv[1])
+
+    with open(json_path) as f:
+        data = json.load(f)
+
+    adapter = TypeAdapter(AnonymizedPayload)
+
+    try:
+        adapter.validate_python(data)
```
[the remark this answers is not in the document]
himdel: This seems valid, unexpected keys should be a problem too

metrics_utility/anonymized_rollups/anonymized_rollups.py
himdel: sounds good in principle, except `flatten_json_report` changes things too much at least for some of these (renames, aggregation, extra fields..), so this would have to be a second set of types .. stretch goal for the non-trivial ones? :)

(for the ones that would stay as `: Any`, maybe we don't have to add the type at all? .. the return type is the important one here)

metrics_utility/anonymized_rollups/types.py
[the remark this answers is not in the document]
himdel: the comment might have to be good enough for now,
seems like conditional fields are not so well supported in openapi tooling

himdel: and and if it lets you without blowing up, `rollup_period_collected_events_total` is also just `NotRequired[int]` in practice

(unless there's an empty array of events but a non-zero event count, it will always be missing or int .. but if the typing system complains about that change, it's fine as is)

tools/generate_segment_schema.py
```diff
@@ -0,0 +1,51 @@
...
+
+    defs = schema.pop('$defs', {})
+    openapi = {
+        'openapi': '3.1.0',
+        'info': {'title': 'Segment Anonymized Payload', 'version': '1.0.0'},
+        'paths': {},
+        'components': {
+            'schemas': {
+                'AnonymizedPayload': schema,
+                **defs,
```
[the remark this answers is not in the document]
himdel: Yeah, if the resulting openapi schema allows unkown fields, that's probably a bug
(unless we end up using something like that for in-development fields)

tools/validate_segment_contract.py
```diff
@@ -0,0 +1,50 @@
...
+
+    json_path = Path(sys.argv[1])
+
+    with open(json_path) as f:
+        data = json.load(f)
+
+    adapter = TypeAdapter(StrictPayload)
+
+    try:
+        adapter.validate_python(data, strict=True)
```
AparnaKarve, line 34: Due to `strict=True`, a collection with zero matching jobs produces `int(0)` for duration fields (please verify that). Strict mode would reject int where float is declared.

Changing the duration fields to `int | float` would let the schema accept both 0 and 0.0.

AparnaKarve (approved):

himdel (approved):
Works for me :)

(Reminder - this still needs (as separate PRs (or issues?)):
* the metrics-service part
* wiring into automation
* typechecking the code that's expected to be covered by these types (hm .. maybe .. we might be hitting some python limitations with that)
)

PR #512, opened 2026-07-30
group dependabot updates into single PRs

himdel:
Based on https://github.blog/security/supply-chain-security/tame-dependabot-group-your-updates-slow-the-cadence-keep-security-fast/

Turns out, dependabot can we convinced to do one big update PR .. trying that..

(Not adding an explicit cooldown block — the default 3-day cooldown (`cooldown.default-days: 3`) is already the built-in behavior.)

Also adds `go.mod` update for the mock tools,
and fixes the containers to use hub mirrors too.

+ https://github.com/ansible/metrics-service/pull/376

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,38 @@ updates:
     directory: '/'
     schedule:
       interval: 'monthly'
+    groups:
+      monthly-batch:
+        patterns:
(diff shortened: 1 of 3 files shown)


MilanPospisil (approved):
LGTM 👍

PR #531, opened 2026-08-05
Add update all script into tools

MilanPospisil:
Small tool for updating (cloning or pulling) all relevant repos for emerging services team.

TODO - we may find different place to it in future, maybe handbook? Repository with our team tools including AI agents?


himdel (changes requested):
Has to be a private repo if you add all the gitlab URLs...

what about https://github.com/ansible/emerging-services-utilities ?

PR #530, opened 2026-08-04
AAP-85766 | fix: parameterize host_metric SQL and use multi-column keyset pagination

hunterkepley:
Replace f-string interpolation of the since timestamp and the previous batch's hostname/host_id marker with bound query parameters, and switch the keyset pagination to a proper multi-column comparison matching an ORDER BY hostname, host_id. The LIMIT is bound as a parameter as well.

Assisted by: Claude Code

[AAP-85766]
Dont forget to link back issue to PR.

> [!WARNING]
> **Failure to notify downstream stakeholders (such as the Analytics team) of schema or version changes can result in outages and loss of revenue.**
>
> Please notify stakeholders to disseminate the change correctly, most notably the **Analytics HCC service**.
> A critical ticket to track the change prior to promotion.
> The change should not be merged until dependencies have been updated.

## Description

This parameterizes a SQL statement, as per request of the controller team, to strengthen it

## Required Actions

- [ ] Requires documentation updates

- [ ] Requires downstream repository changes

- [ ] Requires infrastructure/deployment changes

- [ ] Requires coordination with other teams

- [ ] Blocked by PR/MR: #XXX

### Code Quality

- [X] Code is properly linted and formatted.
- [X] All tests (existing and new) pass successfully.
- [X] Tests are added or updated as needed.
- [X] Code includes relevant comments for complex sections.
- [ ] Changes are reviewed and approved by at least two team members.
- [ ] Documentation is updated where applicable.  
       - If updates are required in the [handbook](https://github.com/ansible/handbook), create a separate PR for the handbook repo.

## Notes for Reviewers

Add any additional context or instructions for reviewers here - for example screenshots if needed.

PR #536, opened 2026-08-10
move host_metric SQL into a library collector

himdel:
Issue: AAP-89148

Extract the host_metric query from ExtractorControllerDB into a proper main_hostmetric library collector, keeping the parameterized SQL and keyset pagination from #530. The keyset marker and LIMIT are bound params; since/until use date_where, matching the other collectors.

The collector paginates via a new generic sql_keyset() on the output adapters: DataframeOutput loops keyset pages and concatenates, while CollectionOutput streams the whole result via a single COPY (used by the new main_hostmetric gather collector). ExtractorControllerDB becomes a thin adapter delegating to the collector for the Renewal Guidance report.

---

this allows us to make the source data for renewal guidance report available to metrics-service, cc @D4D3VD4V3
(pending the resolution of `ensure_functions` - relevant to more collectors, https://github.com/ansible/metrics-utility/pull/536#discussion_r3749968250 )

### DIFF
--- a/metrics_utility/automation_controller_billing/extract/extractor_controller_db.py
+++ b/metrics_utility/automation_controller_billing/extract/extractor_controller_db.py
@@ -2,16 +2,17 @@
 
 import datetime
 
-import pandas as pd
-
 from django.db import connection
+
+from metrics_utility.library.collectors.controller.main_hostmetric import main_hostmetric
 
 
 class ExtractorControllerDB:
     """Extracts host_metric data from the AWX/Controller PostgreSQL database.
 
-    Uses marker-based keyset pagination to stream large result sets in batches
-    of :attr:`limit` rows.
+    Thin adapter over the :func:`~metrics_utility.library.collectors.controller.main_hostmetric.main_hostmetric`
+    library collector, which owns the SQL, the custom PostgreSQL helper functions,
+    and the keyset pagination.
     """
 
     def __init__(self, extra_params):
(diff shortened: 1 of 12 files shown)


metrics_utility/library/collectors/controller/main_hostmetric.py
```diff
@@ -0,0 +1,113 @@
...
+            main_hostmetric.automated_counter,
+            main_hostmetric.deleted_counter,
+            main_hostmetric.last_deleted,
+            main_hostmetric.deleted,
+            main_host.ansible_facts->>'ansible_product_serial'::TEXT AS ansible_product_serial,
+            main_host.ansible_facts->>'ansible_machine_id'::TEXT AS ansible_machine_id,
+            CASE
+                WHEN (metrics_utility_is_valid_json(main_host.variables))
+                    THEN main_host.variables::jsonb->>'ansible_host'
+                ELSE metrics_utility_parse_yaml_field(main_host.variables, 'ansible_host' )
```
himdel, line 64: won't be available in service without ansible/awx#16458,
or changes

himdel: Collectors with the `ensure_functions` problem:

* `main_host`
* `main_hostmetric`
* `job_host_summary` - but `job_host_summary_service` is OK

(not a problem for this PR, but can't be enabled in metrics-service before solving)

metrics_utility/automation_controller_billing/extract/extractor_controller_db.py
```diff
@@ -25,160 +26,19 @@ def __init__(self, extra_params):
...
-    @staticmethod
-    def limit():
-        """Return the maximum number of rows fetched per page.
-
-        Returns:
-            int row-count limit.
-        """
-        return 10000
+        if host_metric is not None and not host_metric.empty:
+            yield {'host_metric': host_metric}
```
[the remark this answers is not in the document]
himdel: Does not apply here.

Previously the dataframe engine would just concatenate all the batches anyway, so we're just moving the concat closer to sql.

D4D3VD4V3 (approved):
Looks great to me, Martin!

PR #564, opened 2026-08-27
bump sqlparse to >=0.6.0

ShaiahWren:
[AAP-89829]

In order to address this CVE,

- **What is being changed?** sqlparse is re-resolved to 0.6.0 in uv.lock
- **Why is this change needed?** due to security issues raised by CVE-2026-54284, CVE-2026-59893, and CVE-2026-71491. These are all fixed in 0.6.0,
- **How does this change address the issue?** uv lock --upgrade-package sqlparse forces re-resolution or sqlparse

### Code Quality

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1617,11 +1617,11 @@ wheels = [
 
 [[package]]
 name = "sqlparse"
-version = "0.5.5"
+version = "0.6.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


msmagnanijr (approved):

himdel (commented):
We don't use `sqlparse` directly and this is adding an explicit dependency?

Shouldn't it just be a uv.lock update instead?

himdel (approved):

